PR #13, opened 2021-12-15
Support running SG without create role permissions

caugustus-sourcegraph:
Matches the changes for frontend and prometheus from the https://sourcegraph.com/github.com/sourcegraph/deploy-sourcegraph/-/tree/overlays/non-privileged overlay. Prometheus and Frontend bind to a pre-existing clusterrole view instead of creating their own role.

Not enabled by default.

### DIFF
--- a/sourcegraph/templates/prometheus/prometheus.ConfigMap.yaml
+++ b/sourcegraph/templates/prometheus/prometheus.ConfigMap.yaml
@@ -42,6 +42,7 @@ data:
     # default named port `https`. This works for single API server deployments as
     # well as HA API server deployments.
     scrape_configs:
+    {{- if .Values.prometheus.privileged }}
     - job_name: 'kubernetes-apiservers'
 
       kubernetes_sd_configs:
@@ -111,6 +112,7 @@ data:
         regex: (.+)
         target_label: __metrics_path__
         replacement: /api/v1/nodes/${1}/proxy/metrics
+    {{- end }} # End of privileged config
 
     # Scrape config for service endpoints.
     #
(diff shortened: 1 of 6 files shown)


sourcegraph/templates/prometheus/prometheus.ConfigMap.yaml
```diff
@@ -236,7 +241,7 @@ data:
       # target_label: kubernetes_namespace
       - source_labels: [__meta_kubernetes_namespace]
         action: replace
-        target_label: {{ .Release.Namespace }}
+        target_label: ns
```
kevinwojo, line 244: Was this intended?

sourcegraph/templates/prometheus/prometheus.RoleBinding.yaml
```diff
@@ -0,0 +1,18 @@
+{{- if and .Values.prometheus.enabled (not .Values.prometheus.privileged) -}}
```
kevinwojo, line 1: I was reading through flow-control for helm templates and I could _NOT_ see any negation! Good to know!

sourcegraph/templates/prometheus/prometheus.ConfigMap.yaml
caugustus-sourcegraph: Ah! Yes, this fixes an earlier mistake on my part - this should never have been a dynamic value, just a label name.

kevinwojo: Actually I think I see here ... `target_label` not the value...

kevinwojo (approved):
LGTM

PR #17, opened 2021-12-16
A minikube/kind compatible Values.yaml file

kevinwojo:
I'm going to float this out there to see if you love or hate it 😆 
--

This took inspiration from the minikube overlay over in sourcegraph/deploy-sourcegraph. We do not set the requested resources or resource limits which allows this to run on smaller hardware without having to predict what resources are available to a developer.

Introduces a new variable called `localDevMode` which will remove the resource stanza from all deployments. This is what the minikube overlay does.

**Sample of insufficient resources on laptop**
Screen Shot 2021-12-16 at 4 21 47 PM

**Output of dev mode:** _notice the absence of "resource:" blocks_
[sample-dev-mode.txt](https://github.com/sourcegraph/deploy-sourcegraph-helm/files/7730770/sample-dev-mode.txt)

**Output of non-dev mode:** _This is the default_
[sample-not-dev-mode.txt](https://github.com/sourcegraph/deploy-sourcegraph-helm/files/7730771/sample-not-dev-mode.txt)


sourcegraph/values.yaml
```diff
@@ -8,6 +8,7 @@ sourcegraph:
   imagePullSecrets: []
   labels: {}
   revisionHistoryLimit: 10
+  localDevMode: false # When true, remove all resource stanzas, allowing the scheduler to best-fit pods. Intended for local development with limited resources.
```
caugustus-sourcegraph, line 11: I like it! We could achieve the same thing by providing an override file that sets all the resources to empty, but this config setting will let us build in other local-specific features.

share/local_dev_values.yaml
```diff
@@ -0,0 +1,408 @@
+### Common Configuration ###
```
caugustus-sourcegraph, line 1: What's your thinking with this file? I'm not sure what it's trying to do here. If there are specific settings that need to be changed for local dev, we should either build them into the templates or limit this file to only the overrides.

kevinwojo: I guess maybe I was stuck in a mode of thinking that we provide examples to our customers ... I think it's best if we left this out.

caugustus-sourcegraph (approved):

PR #23, opened 2022-01-03
tests: add buildkite pipeline and helm-unittest

kevinwojo:


scripts/ci/validate-schema.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+
```
jhchabran, line 3: `set -euf -o pipefail` is missing in all scripts 😊  Otherwise, if a command fails, the script will continue to run!

scripts/ci/validate-schema.sh
```diff
@@ -0,0 +1,11 @@
...
+# Install kubeval
```
caugustus-sourcegraph, line 4: When I looked at this a few months back, I ended up settling on kubeconform instead of kubeval - https://github.com/sourcegraph/deploy-sourcegraph/commits/caug/add-kubeval. Can't remember exactly why, maybe because it's the same person who's maintaining that forked schema?

kevinwojo: Yes you're correct! kubeval died apparently and kubeconform took the spot. Will switch over.

sourcegraph/tests/localDevMode_test.yaml
```diff
@@ -0,0 +1,17 @@
+suite: localDevMode
+templates:
+  - frontend/sourcegraph-frontend.Deployment.yaml
+tests:
+  - it: should not have a resource block when localDevMode=true
```
caugustus-sourcegraph, line 5: This could get quite verbose for extended tests, but it's readable and easy to expand. It would be ideal if we could pass in an entire values.yaml instead of specifying keys directly here, but https://github.com/quintush/helm-unittest/issues/132 is unanswered so maybe that means we can't.

The [snapshot testing feature](https://github.com/quintush/helm-unittest#snapshot-testing) is interesting - it might be a way we can get a more comprehensive test for "free" (the testing would be more fragile against intentional changes).

kevinwojo: I think this is what we're needing for more complex tests?

``` 
 -v, --values stringArray     absolute or glob paths of values files location, default no values files
```

PR #24, opened 2022-01-03
Add Github Action to publish chart versions

caugustus-sourcegraph:
Uses https://github.com/marketplace/actions/helm-chart-releaser to auto-publish the helm chart to a Github page (https://sourcegraph.github.io/deploy-sourcegraph-helm/). Right now it will publish any commit on main to the releases page, but we can make it dependent on checks passing (like those added in https://github.com/sourcegraph/deploy-sourcegraph-helm/pull/23).

### DIFF
--- a/.github/workflows/chart_publish.yml
+++ b/.github/workflows/chart_publish.yml
@@ -1,0 +1,30 @@
+name: Release Charts
+
+on:
+  push:
+    branches:
+      - main
+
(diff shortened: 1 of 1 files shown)


kevinwojo (approved):
This is way cool. We'll likely need to iron out some bumps on the CI side but excited to have CI and a release process in-place.

PR #38, opened 2022-01-20
Add mandatory /tmp dir to searcher deployment, fix missing volume configuration

caugustus-sourcegraph:
The searcher deployment relies on having a writable `/tmp` directory. Since we run by default with a read-only root file system, structural searches fail to run without this directory.

While adding the /tmp mount, I noticed that the optional volume configuration was missing on many deployments, so I fixed that as well.

### DIFF
--- a/charts/sourcegraph/Chart.yaml
+++ b/charts/sourcegraph/Chart.yaml
@@ -5,7 +5,7 @@ description: Chart for installing Sourcegraph
 type: application
 
 # Chart version, separate from Sourcegraph
-version: 0.1.3
+version: 0.1.4
 
 # Version of Sourcegraph release
 appVersion: "3.35.1"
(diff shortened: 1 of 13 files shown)


filiphaftek (approved):
Great finding 👍  One question - should this be also added somewhere in the README? or in the values.yaml (commented out)? so every customer knows straight away how to add tmp volumes without going through the templates? i.e. 
```
- to add extra rw volume for /tmp please setup in values.yaml
<app>:
   extraVolumeMounts:
            < and sample here >
```

caugustus-sourcegraph:
@filiphaftek - Great idea! Added this configuration to the `examples` folder (https://github.com/sourcegraph/deploy-sourcegraph-helm/pull/38/commits/ae9245b936dcabccc9efd9dd5dc4d9665c66e5e9).

PR #39, opened 2022-01-28
Default to latest Sourcegraph release (3.36.2)

caugustus-sourcegraph:
No significant changes for this release, version bump only

### DIFF
--- a/charts/sourcegraph/Chart.yaml
+++ b/charts/sourcegraph/Chart.yaml
@@ -5,7 +5,7 @@ description: Chart for installing Sourcegraph
 type: application
 
 # Chart version, separate from Sourcegraph
-version: 0.1.4
+version: 0.2.0
 
 # Version of Sourcegraph release
-appVersion: "3.35.1"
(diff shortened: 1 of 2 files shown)


kevinwojo (approved):
confirming bump

michaellzc (approved):

michaellzc:
just curious, is this a manual change or it was done by batch change?

caugustus-sourcegraph:
@michaellzc - I ran a local script that pulls the images from the release branches in the deploy repos 😅  I haven't checked it in because it's a temporary hack and not for anyone to use.

This project is still in beta and hasn't yet been incorporated into the batch changes functionality. The tricky part is that the batch changes ultimately rely on https://github.com/sourcegraph/update-docker-tags, which only works against full image references. The values.yaml doesn't expose the full image directly so we need to make some code changes to support the new use case. That work isn't yet planned but it will be required soon.

filiphaftek (approved):
👍

PR #41, opened 2022-02-18
Add migrator initContainers to frontend

michaellzc:
close https://github.com/sourcegraph/sourcegraph/issues/31257

Not blocking 3.37 release

### Should we use a `Job` with [helm hook](https://helm.sh/docs/topics/charts_hooks/)?

I don't have good experience working with helm hook in the past and often yield unexpected result. For example, the `helm install/update` will just fail when the cluster fails to schedule pods, or `pgsql` is not ready yet.

On the other hand, we can run the `Job` with `post-install` and `post-upgrade` hooks. Maybe it can help increase visibility when there's a failing migration?

### DIFF
--- a/charts/sourcegraph/templates/frontend/sourcegraph-frontend.Deployment.yaml
+++ b/charts/sourcegraph/templates/frontend/sourcegraph-frontend.Deployment.yaml
@@ -45,6 +45,26 @@ spec:
         app: sourcegraph-frontend
         deploy: sourcegraph
     spec:
+      initContainers:
+      {{- if .Values.migrator.enabled }}
+      - name: migrator
+        image: {{ include "sourcegraph.image" (list . "migrator") }}
+        args:
+        - up
+        env:
+        {{- range $name, $item := .Values.frontend.env }}
+        - name: {{ $name }}
+          {{- $item | toYaml | nindent 10 }}
+        {{- end }}
+        {{- range $name, $item := .Values.migrator.env }}
+        - name: {{ $name }}
+          {{- $item | toYaml | nindent 10 }}
+        {{- end }}
+        resources:
+          {{- toYaml .Values.migrator.resources | nindent 10 }}
+        securityContext:
+          {{- toYaml .Values.migrator.podSecurityContext | nindent 10 }}
+      {{- end }}
       containers:
       - name: frontend
         image: {{ include "sourcegraph.image" (list . "frontend") }}
--- a/charts/sourcegraph/values.yaml
+++ b/charts/sourcegraph/values.yaml
@@ -221,6 +221,19 @@ frontend:
     create: true
     name: sourcegraph-frontend
 
+migrator:
+  enabled: true
+  image:
+    defaultTag: 3.37.0@sha256:404df69cfee90eaa9a3ab8b540a2d9affd22605caa5326a8ac4ba016e1d6d815
+    name: "migrator"
+  env: {}
+  resources: {}
+  podSecurityContext:
+    allowPrivilegeEscalation: false
+    runAsUser: 100
+    runAsGroup: 101
+    readOnlyRootFilesystem: true
+
 githubProxy:
   image:
     defaultTag: 3.37.0@sha256:3b173e36f958b68479ae829d784c63346701df417afa14d14ae657a84e630dd5


michaellzc:
* main
  * **PR #41** <a href="https://app.graphite.dev/github/pr/sourcegraph/deploy-sourcegraph-helm/41" target="_blank">Graphite</a>  (this PR)

charts/sourcegraph/templates/frontend/sourcegraph-frontend.Deployment.yaml
```diff
@@ -45,6 +45,26 @@ spec:
...
+          {{- $item | toYaml | nindent 10 }}
+        {{- end }}
+        {{- range $name, $item := .Values.migrator.env }}
+        - name: {{ $name }}
+          {{- $item | toYaml | nindent 10 }}
+        {{- end }}
+        resources:
+          {{- toYaml .Values.migrator.resources | nindent 10 }}
+        securityContext:
+          {{- toYaml .Values.migrator.podSecurityContext | nindent 10 }}
```
michaellzc, line 66: I feel like the way we use `podSecurityContext` and `securityContext` values is wrong

`podSecurityContext` should be the pod-level security context

`securityContext` (or `containerSecurityContext`) is the container-leverl security context.

That said, I still following the current convention. If we do decide to change this, it should be work on from a separate PR.

charts/sourcegraph/templates/frontend/sourcegraph-frontend.Deployment.yaml
```diff
@@ -45,6 +45,26 @@ spec:
...
+        - up
+        env:
+        {{- range $name, $item := .Values.frontend.env }}
+        - name: {{ $name }}
+          {{- $item | toYaml | nindent 10 }}
+        {{- end }}
+        {{- range $name, $item := .Values.migrator.env }}
+        - name: {{ $name }}
+          {{- $item | toYaml | nindent 10 }}
+        {{- end }}
```
michaellzc, line 62: I am debating whether reusing all `frontend` env var is a good approach, thoughts?

caugustus-sourcegraph: I like it! Agreed that they're not all required, but it's nice not to have to configure anything extra.

charts/sourcegraph/templates/frontend/sourcegraph-frontend.Deployment.yaml
caugustus-sourcegraph: When I set the chart up I tried to follow the existing convention from the kustomize manifests, but I agree - the distinction is not very clear.

charts/sourcegraph/values.yaml
```diff
@@ -221,6 +221,15 @@ frontend:
     create: true
     name: sourcegraph-frontend
 
+migrator:
+  enabled: true
+  image:
+    defaultTag: insiders@sha256:18fb43c040e91bce06d8123e0f859cdfb1eded897f5c0b3a138864d9a885c1eb
```
caugustus-sourcegraph, line 227: Let's wait to merge this until the 3.37 release is available, and use that as the defaultTag

charts/sourcegraph/values.yaml
```diff
@@ -221,6 +221,15 @@ frontend:
...
+    name: "migrator"
+  env: {}
+  resources: {}
+  podSecurityContext: {}
```
caugustus-sourcegraph, line 231: This should default to the same context as in this file: https://github.com/sourcegraph/deploy-sourcegraph/blob/master/overlays/non-privileged/frontend/sourcegraph-frontend.Deployment.yaml#L10

charts/sourcegraph/templates/frontend/sourcegraph-frontend.Deployment.yaml
```diff
@@ -45,6 +45,26 @@ spec:
         app: sourcegraph-frontend
         deploy: sourcegraph
     spec:
+      initContainers:
+      {{- if .Values.migrator.enabled }}
+      - name: migrator
+        image: {{ include "sourcegraph.image" (list . "migrator") }}
+        args:
+        - up
```
caugustus-sourcegraph, line 53: This should probably be a configurable option

charts/sourcegraph/values.yaml
```diff
@@ -221,6 +221,15 @@ frontend:
     create: true
     name: sourcegraph-frontend
 
+migrator:
```
caugustus-sourcegraph, line 224: I wonder if we should also enable running the migrator as the standalone Job (not associated with helm hooks, just as an option). This would be in addition to the initContainer, and accept a custom arg.

The reason for this would be to make it possible to use the chart to fix database migration errors (https://docs.sourcegraph.com/admin/how-to/dirty_database#3-add-a-migration-log-entry), or to allow customers to deploy the migration job ahead of a general upgrade.

charts/sourcegraph/templates/frontend/sourcegraph-frontend.Deployment.yaml
michaellzc: > This should probably be a configurable option

I was hoping we can make all args or command configurable at once via https://github.com/sourcegraph/sourcegraph/issues/29063

caugustus-sourcegraph: 😆  I forgot that issue existed / hadn't been completed yet

charts/sourcegraph/values.yaml
michaellzc: how about

## Option 1
```yml
migrator:
  type: "" # `initContainer` or `job`
```

Then we render `initContainers` or a new `Job` depending on the values

## Option 2

Create a new chart `sourcegraph/migrator`, by doing so, it makes people's life easier to just run the migration. This could be quite helpful for stuff like manually fixing a dirty migration

```bash
helm upgrade sourcegraph-migrator sourcegraph/migrator --values args=<>
```

or

just render the manifest if the client doesn't want to use helm to update stuff

```bash
helm template sourcegraph-migrator sourcegraph/migrator --values args=<>
```

caugustus-sourcegraph: Option 1 doesn't quite work - we need to be able to run both `initContainer` and `Job` at the same time, and with different args.

Option 2 is interesting. As a standalone chart it's easily uninstalled after the migration runs. It could even be called as an optional dependency of this chart, so that env var's could be reused instead of having to specify them separately. Let's experiment with this approach and see if any surprises come up!

caugustus-sourcegraph (approved):

PR #42, opened 2022-02-18
Add sourcegraph-migrator chart

michaellzc:
This PR added a migrator chart to deploy a single `Job` to execute the `migrator` command.

The `migrator` pod is configured with `spec.template.spec.restartPolicy=Never` and `.spec.backoffLimit=5`, so we don't lose container logs in the event of a failure (otherwise, the pod is restarted instead of recreated).

## Test Plan

```sh
kind create kind
```

Create an `override.yaml` at `charts/sourcegraph/override.yaml`

```yaml
storageClass:
  create: false
  name: standard

sourcegraph:
  localDevMode: true
  image:
    defaultTag: 3.37.0-rc.1
    useGlobalTagAsDefault: true

frontend:
  env:
    NEW_MIGRATIONS: 
      value: "true" # is this still needed?

migrator:
  enabled: false

pgsql:
  enabled: true

cadvisor:
  enabled: false
```

Deploy the parent chart with the override

```sh
helm install -n sourcegraph -f charts/sourcegraph/override.yaml sourcegraph charts/sourcegraph/.
```

Wait for all pods are fully created, some (e.g. frontend) will stay at crashloop, but that's okay

Deploy the migrator

```sh
helm install -n sourcegraph -f charts/sourcegraph/override.yaml sourcegraph-migrator charts/sourcegraph-migrator/.
```

```sh
k rollout restart deployment sourcegraph-frontend
```

;) now you have a working sg


michaellzc:
* main
  * **PR #41** <a href="https://app.graphite.dev/github/pr/sourcegraph/deploy-sourcegraph-helm/41" target="_blank">Graphite</a> 
    * **PR #42** <a href="https://app.graphite.dev/github/pr/sourcegraph/deploy-sourcegraph-helm/42" target="_blank">Graphite</a>  (this PR)

charts/sourcegraph-migrator/templates/migrator/sourcegraph-migrator.Job.yaml
```diff
@@ -0,0 +1,75 @@
...
+  annotations:
+    description: Performs Sourcegraph database migrations.
+  labels:
+    {{- include "sourcegraph.labels" . | nindent 4 }}
+    {{- if .Values.migrator.labels }}
+      {{- toYaml .Values.migrator.labels | nindent 4 }}
+    {{- end }}
+    deploy: sourcegraph
+    app.kubernetes.io/component: migrator
+  name: {{ default "migrator" .Values.migrator.name }}-{{ randAlphaNum 5 | lower }}
```
michaellzc, line 13: `{{ randAlphaNum 5 | lower }}` ensures user don't have to uninstall the chart every time.

Upon each upgrade, previous job will be delete by helm or user can clean them up manually.

charts/sourcegraph-migrator/templates/migrator/sourcegraph-migrator.Job.yaml
```diff
@@ -0,0 +1,75 @@
...
+      {{- if .Values.migrator.podLabels }}
+      {{- toYaml .Values.migrator.podLabels | nindent 8 }}
+      {{- end }}
+        job: migrator
+        deploy: sourcegraph
+    spec:
+      containers:
+      - name: migrator
+        image: {{ include "sourcegraph.image" (list . "migrator") }}
+        args: {{- default (list "up") .Values.migrator.args | toYaml | nindent 8 }}
```
michaellzc, line 42: address https://github.com/sourcegraph/sourcegraph/issues/29063

charts/sourcegraph-migrator/values.yaml
```diff
@@ -0,0 +1,113 @@
...
+#   # Set disk size on PVC's for stateful deployments
+#   storageSize: ""
+#   # Optionally bind PVC's to an existing PV
+#   volumeName: ""
+#   # Use an existing configmap (specified by name) instead of creating one
+#   existingConfig: ""
+#   # Toggle deployment of applications on/off. Applies to databases and third-party applications
+#   enabled: true
+
+frontend:
```
caugustus-sourcegraph, line 72: The mixture of frontend and migrator env's feels messy here - the migrator really is a standalone job here, not associated with the frontend at all.

michaellzc: @caugustus-sourcegraph 

Yeah, I agree a mix of frontend and migrator values is really confusing.

I will just move all frontend values to migrator, and we can document that clients can reuse the env var for frontend in migrator.

michaellzc:
@caugustus-sourcegraph 

> Same as above with the backofflimit - why prevent this job from retrying?

This is intentional. ~When the pod is restarted, container logs are gone are well. I think it's important for the user to always check the logs of the previous run before running it again.~

I just realize I can mix `spec.template.spec.restartPolicy=Never` and `.spec.backoffLimit=5` (or any number > 1) to tell the job controller to create a new pod for each retry, so we won't lose the container logs.

charts/sourcegraph-migrator/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Run database migration
+
+[`migrator up`](https://docs.sourcegraph.com/admin/how-to/manual_database_migrations#up)
+
+- Perform database migrations prior to upgrading the Sourcegraph deployment
+- Perform initial migrations against external PostgreSQL databases prior to the Sourcegraph deployment
+
+```sh
+helm upgrade --install -f <your-override-file.yaml> --set "migrator.args={up,-db=all}" sg-migrator sourcegraph/sourcegraph-migrator
```
caugustus-sourcegraph, line 40:
```suggestion
helm upgrade --install -f <your-override-file.yaml> sg-migrator sourcegraph/sourcegraph-migrator
```

For most customers the default will work fine here. If not, the permanent arg customization can be set in that override file.

charts/sourcegraph-migrator/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+```sh
+helm repo add sourcegraph https://sourcegraph.github.io/deploy-sourcegraph-helm/
+```
+
+## Usage
+
+> If you are not using external databases, the chart has to be installed in the same namespace as the parent [sourcegraph/sourcegraph] chart
+
+[sourcegraph/sourcegraph-migrator] chart requires the correct `PG*` and `CODEINTEL_PG*` environment variables to be configured at `migrator.env`. Learn more about [using your own PostgreSQL server]. `PG*` and `CODEINTEL_PG*` environment variables are compatible with the `frontend.env` values from the parent [sourcegraph/sourcegraph] chart.
```
caugustus-sourcegraph, line 28: Code Insights too!

caugustus-sourcegraph (approved):
I'm curious how the values propagation works when this is added as a dependency of the main chart. For instance, as a standalone chart the `sourcegraph.labels` helper template will generate different values than the main chart, and in the values file we duplicate configuration found in the main chart (i.e. `sourcegraph.imagePullSecrets`).

Will there be a way to make them consistent or will customers always need to maintain a duplicate set of some configuration settings?

charts/sourcegraph-migrator/README.md
michaellzc: > Code Insights too!

It looks like we have more docs to update!

https://docs.sourcegraph.com/admin/external_services/postgres#instructions

michaellzc:
If we make this a dependency of the parent chart, it could be tricky, cuz user would have to configure the migratory job with

```yaml
migrator: # of something alias
  migrator:
    env: <things>
```

### Optional 1 - avoid nesting

Or, we could avoid nesting `migrator` config under `migrator` in the migrator chart, just keep things at the top level

```yaml
env:
  PG*:
    value: ""
podSecurityContext
```

### Option 2 - some yaml magic

`sourcegraph/Chart.yaml`

```yaml
dependencies:
  - name: sourcegraph-migrator
    version: 0.1.0
    repository: file:///Users/michael/Code/sourcegraph/deploy-sourcegraph-helm/charts/sourcegraph-migrator
    alias: migratorJob
```

We can ask users to use alias to reuse config for migrator

```yaml
migrator: &migrator
  enabled: false
  args: ["up", "-db=frontend"]

migratorJob: *migrator
```

### Option 3

Just don't use migratory as a dependency, but an optional way (the right way) to do database related operation without interfering with the frontend deployment

PR #48, opened 2022-02-25
Fix securityContext values name

michaellzc:
We're currently using and

- `podSecurityContext` as container-level `securityContext` (the naming is misleading)
- `securityContext` as pod-level `securityContext` (okay, but we should be more explicit)

This PR

- renamed `podSecurityContext` to `containerSecurityContext`
- renamed `securityContext` to `podSecurityContext`

## Test plan

The test plan is to `diff` the output of rendered manifest, no change is expected.

```sh
kind create cluster
```

Revert to `main`

```sh
 git checkout HEAD~
```

```sh
helm upgrade --install -n sourcegraph --set storageClass.create=false,storageClass.name=standard,sourcegraph.localDevMode=true sourcegraph charts/sourcegraph/.
```

```sh
git checkout 02-25-Fix_securityContext_values_name
```

```sh
helm diff -n sourcegraph --set storageClass.create=false,storageClass.name=standard,sourcegraph.localDevMode=true sourcegraph charts/sourcegraph/.
```

There should be no difference.

close https://github.com/sourcegraph/sourcegraph/issues/31888

### DIFF
--- a/charts/sourcegraph/CHANGELOG.md
+++ b/charts/sourcegraph/CHANGELOG.md
@@ -1,0 +1,27 @@
+<!-- README:
+Use `**BREAKING**:` to denote a breaking change
+-->
+
+# Changelog
+
+All notable changes to Sourcegraph are documented in this file.
+
+<!-- START CHANGELOG -->
+
+## Unreleased
+
+### Added
+
+- 
+
+### Changed
+
+- **BREAKING**: The use of `.<service>.podSecurityContext` and `.<service>.securityContext` is causing confusion and misleading, we are using `podSecurityContext` as container-level `securityContext` and `securityContext` as pod-level `securityContext`. `podSecurityContext` are renamed to `containerSecurityContext`, and `securityContext` are renamed to `podSecurityContext` to better illustrate their actual usage [#48](https://github.com/sourcegraph/deploy-sourcegraph-helm/pull/48)
+
+### Fixed
+
+-
+
+### Removed
+
+-
--- a/charts/sourcegraph/CHANGELOG.md
+++ b/charts/sourcegraph/CHANGELOG.md
@@ -16,7 +16,7 @@ All notable changes to Sourcegraph are documented in this file.
 
 ### Changed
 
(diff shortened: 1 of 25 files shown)


michaellzc:
* main
  * **PR #48** <a href="https://app.graphite.dev/github/pr/sourcegraph/deploy-sourcegraph-helm/48" target="_blank">Graphite</a>  (this PR)

michaellzc:
> This is a breaking change, so it requires bumping the minor version on the chart version.
> 
> Now that we've started combining multiple changes into a single release, and now have a breaking change, I think we should start producing a changelog. Before, it wasn't very important because one commit == one release.

Added a changelog to record breaking changes. Also, I added a note to this the `RELEASE.md` https://github.com/sourcegraph/deploy-sourcegraph-helm/pull/50/files#diff-2b1b69303b927a484e02c7fad9fc87d0d3ff0dc22ae1da0ecd0dc935d922a23cR7 to keep an eye for breaking changes during release. Assuming we're fine with my proposed release process (multiple commits in one release instead of bumping the chart.yaml in every single commit)

charts/sourcegraph/CHANGELOG.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Unreleased
+
+### Added
+
+- 
+
+### Changed
+
+- **BREAKING**: The use of `.<service>.podSecurityContext` and `.<service>.securityContext` is causing confusion and misleading, we are using `podSecurityContext` as container-level `securityContext` and `securityContext` as pod-level `securityContext`. `podSecurityContext` are renamed to `containerSecurityContext`, and `securityContext` are renamed to `podSecurityContext` to better illustrate their actual usage [#48](https://github.com/sourcegraph/deploy-sourcegraph-helm/pull/48)
```
caugustus-sourcegraph, line 19:
```suggestion
- **BREAKING**: Uses of `podSecurityContext` have been renamed to `containerSecurityContext`, and `securityContext` renamed to `podSecurityContext` to better illustrate their actual usage [#48](https://github.com/sourcegraph/deploy-sourcegraph-helm/pull/48)
```

caugustus-sourcegraph (approved):

PR #52, opened 2022-02-28
Support custom service account name

michaellzc:
This PR enables users to override service account for all deployed resources

### Initial deployment

```sh
kind create cluster
```

`override.yaml`

```yml
storageClass:
  create: false # Disable if you have your own existing storage class
  name: standard

sourcegraph:
  localDevMode: true
```

```sh
helm upgrade --install -n sourcegraph -f charts/sourcegraph/override.yaml sourcegraph charts/sourcegraph/
```

### Enable SA

Updated `override.yaml`

```yml
storageClass:
  create: false # Disable if you have your own existing storage class
  name: standard

sourcegraph:
  localDevMode: true

codeInsightsDB:
  serviceAccount:
    create: true
codeIntelDB:
  serviceAccount:
    create: true
githubProxy:
  serviceAccount:
    create: true
gitserver:
  serviceAccount:
    create: true
indexedSearch:
  serviceAccount:
    create: true
minio:
  serviceAccount:
    create: true
pgsql:
  serviceAccount:
    create: true
preciseCodeIntel:
  serviceAccount:
    create: true
redisCache:
  serviceAccount:
    create: true
redisStore:
  serviceAccount:
    create: true
repoUpdater:
  serviceAccount:
    create: true
searcher:
  serviceAccount:
    create: true
symbols:
  serviceAccount:
    create: true
syntectServer:
  serviceAccount:
    create: true
tracing:
  serviceAccount:
    create: true
worker:
  serviceAccount:
    create: true
```

```sh
 helm diff -n sourcegraph -f charts/sourcegraph/override.yaml sourcegraph charts/sourcegraph/ 
```

- Verify rendered manifest is expected
- Verify sourcegraph server still works

fix https://github.com/sourcegraph/sourcegraph/issues/31889

PR #50, opened 2022-02-26
Revamp config docs with helm-docs

michaellzc:
This PR added autogen docs based on `values.yaml`.

`charts/<chart_name>/README.md` should no longer be edited, and changes should be made in `charts/<chart_name>/README.md.gotmpl`

View the rendered version here

- https://github.com/sourcegraph/deploy-sourcegraph-helm/tree/02-25-Revamp_config_docs_with_helm-docs/charts/sourcegraph
- https://github.com/sourcegraph/deploy-sourcegraph-helm/tree/02-25-Revamp_config_docs_with_helm-docs/charts/sourcegraph-migrator

## Test plan

Run `./scripts/helm-docs.sh`, you should be able to generate the README.md

close https://github.com/sourcegraph/sourcegraph/issues/31016

PR #53, opened 2022-03-01
Add examples for using external redis

michaellzc:
Added examples to show how to use external redis instances

## Test plan

N/A, it's the helm-way of configuring external redis instances

- https://docs.sourcegraph.com/admin/install/kubernetes/configure#configure-custom-redis
- https://docs.sourcegraph.com/admin/external_services/redis#using-your-own-redis-server

fix https://github.com/sourcegraph/sourcegraph/issues/31874


michaellzc:
* main
  * **PR #50** <a href="https://app.graphite.dev/github/pr/sourcegraph/deploy-sourcegraph-helm/50" target="_blank">Graphite</a> 
    * **PR #52** <a href="https://app.graphite.dev/github/pr/sourcegraph/deploy-sourcegraph-helm/52" target="_blank">Graphite</a> 
    * **PR #53** <a href="https://app.graphite.dev/github/pr/sourcegraph/deploy-sourcegraph-helm/53" target="_blank">Graphite</a>  (this PR)

michaellzc:
@mucles maybe you can help take a look? I am hoping your CE experience can help judge if these examples make sense or not

charts/sourcegraph/examples/external-redis/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+The string must either have the format `$HOST:PORT` or follow the [IANA specification for Redis URLs](https://www.iana.org/assignments/uri-schemes/prov/redis) (e.g., redis://:mypassword@host:6379/2)
+
+### `REDIS_STORE_ENDPOINT`
+
+The string must either have the format `$HOST:PORT` or follow the [IANA specification for Redis URLs](https://www.iana.org/assignments/uri-schemes/prov/redis) (e.g., redis://:mypassword@host:6379/2)
+
+## Notes
+
+You may store these sensitive environment variables in a [Secret](https://kubernetes.io/docs/concepts/configuration/secret/).
```
caugustus-sourcegraph, line 32: 🤔  Are redis endpoints considered sensitive info? Doesn't seem like it matters much - if the endpoint is routable you're already in a bad state. Maybe better to show as a ConfigMap example?

caugustus-sourcegraph (approved):

charts/sourcegraph/examples/external-redis/README.md
michaellzc: > 🤔 Are redis endpoints considered sensitive info? Doesn't seem like it matters much - if the endpoint is routable you're already in a bad state. Maybe better to show as a ConfigMap example?

The `ENDPOINT` env var may or may not include credential `redis://somepassword@external-redis/0`, so it could be sensitive.

But good point, we can add a ConfigMap example.

charts/sourcegraph/examples/external-redis/README.md
```diff
@@ -0,0 +1,53 @@
+# Using external Redis instances
+
+Sourcegraph deployment by default ships two separate Redis instances for different purposes
+
+- [redis-cache.Deployment.yaml](../../templates/redis/redis-cache.Deployment.yaml)
+- [redis-store.Deployment.yaml](../../templates/redis/redis-store.Deployment.yaml)
+
+When using external Redis instances, you’ll need specify the corresponding environment variable for [each of the following deployments](https://docs.sourcegraph.com/admin/install/kubernetes/configure#configure-custom-redis)
```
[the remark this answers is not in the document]
michaellzc:
```suggestion
When using external Redis instances, you’ll need specify the corresponding environment variable for [each of the following deployments](https://docs.sourcegraph.com/admin/install/kubernetes/configure#configure-custom-redis)
```

I was thinking explicitly referencing the deployment template in this repo, but my main concern is the list may get out of date. The list on the referenced docsite is kind of auto-gen already https://github.com/sourcegraph/sourcegraph/blob/main/doc/admin/install/kubernetes/configure.md?plain=1#L709

How about we make a list and mention this link may get out of date, and users should always double-check the referenced docsite

charts/sourcegraph/examples/external-redis/override-separate-configmap.yaml
```diff
@@ -0,0 +1,86 @@
+# Demonstrate using external redis instance(s)
```
caugustus-sourcegraph, line 1: What is the benefit of showing the exact same setup, with both a secret and a configmap? There's no significant difference between them, and it means more maintenance in the future.

michaellzc: > What is the benefit of showing the exact same setup, with both a secret and a configmap? There's no significant difference between them, and it means more maintenance in the future.

🧐 I guess adding a reference k8s doc should be sufficient enough

still figuring out the right balance of how much k8s knowledge we should cover in docs/examples.

PR #56, opened 2022-03-02
Add advanced scheduling example override

michaellzc:
This PR added an example override to show advanced scheduling

Also added `tpl` support to affinity rule, so string template can be evaluated as template during rendering

`override.yaml`
```
labels:
  app.kubernetes.io/name: '{{ include "sourcegraph.name" . }}'
  app.kubernetes.io/instance: '{{ .Release.Name }}'
```

https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/280065f45e5738f1a1deed61b235192d1e5893c4/charts/sourcegraph/templates/_helpers.tpl#L75-L86

## Test plan

```sh
helm template -f ./charts/sourcegraph/examples/advanced-scheduling/override.yaml sourcegraph charts/sourcegraph/.
```

close https://github.com/sourcegraph/sourcegraph/issues/31875

### DIFF
--- a/charts/sourcegraph/examples/advanced-scheduling/override.yaml
+++ b/charts/sourcegraph/examples/advanced-scheduling/override.yaml
@@ -2,7 +2,7 @@
 # - the use of advanced scheduling in Kubernetes, including
 #   - node selector
 #   - tolerations
-#   - affinity to scatter services across multi availability zones 
+#   - affinity to scatter services across multiple nodes
 # - the use of YAMl anchor and alias to keep your override.yaml DRY
 #
 # For scaling guide, you should refer to our official docs 
@@ -31,7 +31,7 @@ frontend:
     podAntiAffinity:
       preferredDuringSchedulingIgnoredDuringExecution:
         - podAffinityTerm:
-            topologyKey: topology.kubernetes.io/zone
+            topologyKey: kubernetes.io/hostname
             labelSelector:
               matchLabels:
                 <<: *commonSelectorLabels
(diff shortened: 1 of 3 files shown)


charts/sourcegraph/examples/advanced-scheduling/override.yaml
```diff
@@ -0,0 +1,108 @@
...
+    value: "some-value"
+    effect: "NoSchedule"
+
+# Common select labels to select pods that belong to the current helm release
+selectorLabels: &commonSelectorLabels
+  app.kubernetes.io/name: '{{ include "sourcegraph.name" . }}'
+  app.kubernetes.io/instance: '{{ .Release.Name }}'
+
+frontend:
+  replicaCount: 10
```
michaellzc, line 24: these are just arbitrary number

charts/sourcegraph/examples/advanced-scheduling/override.yaml
```diff
@@ -0,0 +1,110 @@
...
+              matchLabels:
+                <<: *commonSelectorLabels
+                app: symbols
+          weight: 100
+
+# TODO
+# - are they horizontally scalable?
+repoUpdater: {}
+syntectServer: {}
+worker: {}
```
michaellzc, line 110: If they're not scalable, should we remove the `replicaCount` for these services as well? Similar to redis and pg

caugustus-sourcegraph (approved):
🙌 for tests!

My only hesitation with this example is that we know there are performance costs when going multi-AZ (https://github.com/sourcegraph/sourcegraph/issues/28658 will attempt to measure the impact).

michaellzc:
> My only hesitation with this example is that we know there are performance costs when going multi-AZ ([sourcegraph/sourcegraph#28658](https://github.com/sourcegraph/sourcegraph/issues/28658) will attempt to measure the impact).

Good point.

NFS/EFS is a horrible idea. EBS doesn't support multi-zone, and k8s has no way to always schedule a pod to a specific zone where the PV is provisioned. Too bad `WaitForFirstConsumer` only helps during provisioning time to ensure pod and PV are created in the same zone.

I am going to update the example to use the hostname as the topology key.

PR #57, opened 2022-03-02
Refactor postgres reference

david-sandy:
Updated to new postgres-12-alpine image.


charts/sourcegraph/values.yaml
```diff
@@ -555,12 +555,8 @@ pgsql:
...
   existingConfig: "" # Name of an existing configmap
   image:
-    # -- Docker image tag for the `pgsql` image
-    defaultTag: 3.37.0@sha256:af00a4465f435742df9980cc9d8a8bda49d6adea173558a371e276b476ea94ae
-    # -- Docker image name for the `pgsql` image
-    name: "postgres-12.6-alpine"
-  # -- Security context for the `pgsql` container,
-  # learn more from the [Kubernetes documentation](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-container)
+    defaultTag: 135107_2022-03-03_9498a8bd3366@sha256:e26b159dc7c0c47d136886390c899816e669a3c2c1ead689bdad0b610364e45e
+    name: "postgres-12-alpine"
```
michaellzc, line 559: We use the comments like the inline comments to document these values and autogen our docs

```
# -- Docker image tag for the `pgsql` image
```

Would you add them back?

michaellzc (changes requested):
we usually only release changes to images and services during the monthly release or patch release.

we release new features to `helm` outside of the monthly release schedule, but we don't change any underlying images.

@caugustus-sourcegraph maybe we should leave this PR open until `3.38` release? Also, we need to figure out a process for `helm` to address changes like this in between monthly releases.

caugustus-sourcegraph:
@michaellzc - Hm, good point about the process. I wonder if the release tooling / batch changes supports adding onto an existing branch, instead of creating a new one. If it does, we could put changes like this on the publish branch so they're not picked up until the release.

For now, I agree that leaving the PR open is the best bet, and we'll merge it in right before the release? 🤷

michaellzc:
> @caugustus-sourcegraph @michaellzc I'm assuming that this just needs to be merged before the release tomorrow, and not the release cut today. Is that correct?

https://github.com/sourcegraph/sourcegraph/issues/31768

at this step `From the [release batch change](https://k8s.sgdev.org/organizations/sourcegraph/batch-changes), merge the release-publishing PRs created previously.`, the release tool will create a batch change that includes a PR to this repository to bump images tags and chart version.

I would just merge these two release-blocker PRs __BEFORE__ you run the `yarn release release:stage` command

caugustus-sourcegraph (approved):

PR #58, opened 2022-03-03
sourcegraph: release 0.4.0

michaellzc:
I backfilled our changelog with commits that I think it's worth mentioning ` git log --oneline -- charts/sourcegraph/`

## Test plan

N/A, new release

### DIFF
--- a/charts/sourcegraph/CHANGELOG.md
+++ b/charts/sourcegraph/CHANGELOG.md
@@ -26,7 +26,7 @@ All notable changes to Sourcegraph are documented in this file.
 
 -
 
-## 1.0.0
+## 0.4.0
 
 :bulb: We revamped the documentation of helm chart configuration options, [learn more](https://github.com/sourcegraph/deploy-sourcegraph-helm/tree/main/charts/sourcegraph#configuration-options)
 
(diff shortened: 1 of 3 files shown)


charts/sourcegraph/Chart.yaml
```diff
@@ -5,7 +5,7 @@ description: Chart for installing Sourcegraph
 type: application
 
 # Chart version, separate from Sourcegraph
-version: 0.3.0
+version: 1.0.0
```
caugustus-sourcegraph, line 8: I'm not sure we're ready to 1.0 yet! Maybe just bump to `0.4.0` and we can 1.0 as part of going GA?

michaellzc, line 8:
```suggestion
version: 0.4.0
```

caugustus-sourcegraph: I know that's not following strict semver, but 1.0 is special 😄

caugustus-sourcegraph (approved):

charts/sourcegraph/CHANGELOG.md
```diff
@@ -25,3 +25,23 @@ All notable changes to Sourcegraph are documented in this file.
 ### Removed
 
 -
+
+## 1.0.0
```
michaellzc, line 29:
```suggestion
## 0.4.0
```

PR #60, opened 2022-03-07
Set default resource requests/limits on the migrator container

caugustus-sourcegraph:
Same as https://github.com/sourcegraph/deploy-sourcegraph/pull/4095 but for helm.

### Test plan

Diffed the rendered manifests and confirmed they look the same as the deploy-sourcegraph PR.


michaellzc (approved):
lgtm

michaellzc:
Actually, need to run `./scripts/helm-docs.sh` and commit the changes

caugustus-sourcegraph:
@michaellzc - aagh, forgot that changing defaults would mean we need to update the docs! Luckily you / the build failures saved me.

PR #65, opened 2022-03-09
Add docs to resource name override

michaellzc:
`<serviceName>.name` doesn't only override deploy/sts name, but also all related resources, e.g. svc, configmap.

I am debating whether this is actually a good idea because renaming `svc` breaks service discovery. This would have been easier if our application code is not relying on the default service name but explicitly using an environment variable or config file to tell what the service name is.

A good example would be redis https://docs.sourcegraph.com/admin/install/kubernetes/configure#configure-custom-redis
changing redis host name require adding environment variables to more than one deployment and none of these deployment use to require a `REDIS_*_ENDPOINT` environment variable to work. 

Also, the way to figure out the dependency chain just feels wonky to me

- https://github.com/sourcegraph/sourcegraph/blob/c4d4caa65bab8480394cfd116b7dc979782050c3/doc/admin/install/kubernetes/configure.md?plain=1#L709-L716
- https://github.com/sourcegraph/sourcegraph/tree/main/dev/depgraph

From a customer perspective, if they choose to override the `.name` values, they will have to update the environment variables across services, and there's a good chance they may get it wrong.

## Option 1

Revise the template to make changing `<serviceName>.name` only affects the deploy/sts name

## Option 2

Provide a sample override to show what env vars have to be added and changed to various deploy/sts. `docker-compose` has done a good job at doing so, but I am not confident we will be able to keep the list up-to-date.

for example, `SEARCHER_URL`, `REPO_UPDATER_URL`, and more

https://github.com/sourcegraph/deploy-sourcegraph-docker/blob/master/docker-compose/docker-compose.yaml

### Checklist

### Test plan

`override.yaml`
```
storageClass:
  create: false
  name: standard

sourcegraph: 
  localDevMode: true
```

Clean install from the latest release

```sh
helm upgrade --install -n sourcegraph -f ./override.yaml sourcegraph sourcegraph/sourcegraph
```

Inspect the diff

```sh
helm diff -n sourcegraph -f ./override.yaml sourcegraph charts/sourcegraph/.
```

close https://github.com/sourcegraph/sourcegraph/issues/32214

### DIFF
--- a/charts/sourcegraph/README.md
+++ b/charts/sourcegraph/README.md
@@ -80,6 +80,7 @@ In addition to the documented values, all services also support the following va
 - `<serviceName>.podSecurityContext` - [learn more](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-pod)
 - `<serviceName>.args` - override default container args
 - `<serviceName>.env` - consult `values.yaml` file
+- `<serviceName>.name` - override the default resource name, e.g. you want add custom prefix or suffix to the resource name.
 
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
--- a/charts/sourcegraph/README.md.gotmpl
+++ b/charts/sourcegraph/README.md.gotmpl
@@ -80,6 +80,7 @@ In addition to the documented values, all services also support the following va
 - `<serviceName>.podSecurityContext` - [learn more](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-pod)
 - `<serviceName>.args` - override default container args
 - `<serviceName>.env` - consult `values.yaml` file
+- `<serviceName>.name` - override the default resource name, e.g. you want add custom prefix or suffix to the resource name.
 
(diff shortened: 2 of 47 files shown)


michaellzc:
* main
  * **PR #65** <a href="https://app.graphite.dev/github/pr/sourcegraph/deploy-sourcegraph-helm/65" target="_blank">Graphite</a>  (this PR)

michaellzc:
I prefer option 1.

AFAIK, https://github.com/sourcegraph/accounts/issues/574 is the only customizing resource name, and they're only adding a suffix to deploy/sts, not service names.

It's just too much of a maintenance burden with pption 2

caugustus-sourcegraph:
@michaellzc - agreed, option 1 sounds like the safest way forward

michaellzc:
@caugustus-sourcegraph MEGA PR sorry... I am not sure if there's a better way to break it down to smaller commits that is reviewable.

caugustus-sourcegraph (approved):
Technically this is a breaking change if anyone has customized names, but I'm ok with not really labeling it like that.

PR #66, opened 2022-03-09
Set deploy_type env var to track helm chart usage

caugustus-sourcegraph:
Enable tracking of helm chart usage by using the deploy_type env var.

### Checklist

### Test plan

Compared diffs before and after - new env var is set. Frontend will crash unless the changes from https://github.com/sourcegraph/sourcegraph/pull/32407 are present.

### DIFF
--- a/charts/sourcegraph/values.yaml
+++ b/charts/sourcegraph/values.yaml
@@ -240,6 +240,8 @@ frontend:
       value: disable
     CODEINTEL_PGUSER:
       value: sg
+    DEPLOY_TYPE:
+      value: helm
     GRAFANA_SERVER_URL:
       value: http://grafana:30070
     JAEGER_SERVER_URL:


caugustus-sourcegraph:
@michaellzc - this is another one for the "can't merge until the release" bucket. I added the `release-blocker` label to this and the postgres PR so we can at least track it until we have a better process.

michaellzc (approved):
lgtm

michaellzc:
Since now we have one more PR that can only be merged right before the release.

Should we add a release label `release-3.39` on them so it will be easier to track? Just as a reminder for our self to merge them before the minor release

camdencheek:
@caugustus-sourcegraph @michaellzc Same question as [here](https://github.com/sourcegraph/deploy-sourcegraph-helm/pull/57#issuecomment-1071149353): this will just be merged before the release tomorrow, right?

PR #69, opened 2022-03-11
Add gcp example to expose sourcegraph publically

michaellzc:
### Checklist


michaellzc:
* main
  * **PR #69** <a href="https://app.graphite.dev/github/pr/sourcegraph/deploy-sourcegraph-helm/69" target="_blank">Graphite</a>  (this PR)

michaellzc:
> LGTM on this approach. Looks like we are essentially adding more docs here as this in the `example` folder

docsite should be the single source of truth, but we also need a place to store all these sample override files and some local context is also nice. I am still working out the balance about what to keep in the helm repo and what go to the docsite, here's something WIP https://github.com/sourcegraph/sourcegraph/pull/32463.

The plan for helm page on docsite is to keep it as a one-pager, so we can break it up or move sections around when the new docsite IA is ready (an initiativeive led by Jeff, Director of Technical Writing). Maybe we will move all example to docsite or figure out a way to reference them from docsite after GA.

PR #75, opened 2022-03-16
Fix tracing partial template

michaellzc:
The `{{- end }}` was misplaced in the partial template.

When `tracingAgent` is disabled, it will leave a dangling `securityContext:` in the rendered manifests. Tracing is always enabled by default, so we missed it

### Checklist

### Test plan

Run `helm template` with `tracingAgent.enabled=false`, the rendered manifest should look good

### DIFF
--- a/charts/sourcegraph/templates/_tracing.tpl
+++ b/charts/sourcegraph/templates/_tracing.tpl
@@ -29,7 +29,7 @@ Define the tracing sidecar
   args:
     - --reporter.grpc.host-port={{ default "jaeger-collector" .Values.tracing.collector.name }}:14250
     - --reporter.type=grpc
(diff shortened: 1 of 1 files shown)


caugustus-sourcegraph (approved):
So many little mistakes caused by not having any tests from the get-go 😄  Thanks for doing all this cleanup

michaellzc:
> So many little mistakes caused by not having any tests from the get-go 😄 Thanks for doing all this cleanup

@caugustus-sourcegraph 

I am trying to recreate dogfood from our helm chart https://github.com/sourcegraph/deploy-sourcegraph-dogfood-k8s/pull/4018

it's a good trial run before GA :)

PR #77, opened 2022-03-18
Support extending postgres server conf

michaellzc:
We can now extend existing postgres configmap instead of requiring maintaining a fork of it

### Checklist

### Test plan

See added unit tests for template testing and sanity check

To check it actually works

```yaml
storageClass:
  create: false
  name: standard
  reclaimPolicy: Delete

sourcegraph: 
  localDevMode: true

pgsql:
  additionalConfig: |
    max_connections = 500
```

Deploy the chart

```sh
kind create cluster
helm upgrade --install -n sourcegraph -f ./override.yaml sourcegraph charts/sourcegraph/.
```

```sh
kubectl exec pgsql-0 -- psql postgres://sg@localhost -c 'show max_connections;'
```

you should see

```
 max_connections
-----------------
 500
(1 row)
```

The default is

https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/646b48eb80456c53a97b83121322c130680847cf/charts/sourcegraph/templates/pgsql/pgsql.ConfigMap.yaml#L76-L78

hence we can tell stuff at the bottom has higher precendence.

close https://github.com/sourcegraph/sourcegraph/issues/32700

### DIFF
--- a/charts/sourcegraph/templates/codeinsights-db/codeinsights-db.ConfigMap.yaml
+++ b/charts/sourcegraph/templates/codeinsights-db/codeinsights-db.ConfigMap.yaml
@@ -1,3 +1,6 @@
+{{- if and .Values.codeInsightsDB.enabled (and .Values.codeInsightsDB.existingConfig .Values.codeInsightsDB.additionalConfig) -}}
+{{- fail "You can only define one of 'codeInsightsDB.existingConfig' and 'codeInsightsDB.additionalConfig' at a time" }}
+{{- end }}
 {{- if and .Values.codeInsightsDB.enabled (not .Values.codeInsightsDB.existingConfig) -}}
 apiVersion: v1
 kind: ConfigMap
@@ -771,4 +774,6 @@ data:
     timescaledb.max_background_workers = 8
     timescaledb.last_tuned = '2021-02-16T03:10:41Z'
     timescaledb.last_tuned_version = '0.10.0'
+
+    {{ tpl .Values.codeInsightsDB.additionalConfig . | nindent 4 | trim }}
 {{- end }}
--- a/charts/sourcegraph/templates/pgsql/pgsql.ConfigMap.yaml
+++ b/charts/sourcegraph/templates/pgsql/pgsql.ConfigMap.yaml
@@ -1,3 +1,6 @@
+{{- if and .Values.pgsql.enabled (and .Values.pgsql.existingConfig .Values.pgsql.additionalConfig) -}}
+{{- fail "You can only define one of 'pgsql.existingConfig' and 'pgsql.additionalConfig' at a time" }}
+{{- end }}
 {{- if and .Values.pgsql.enabled (not .Values.pgsql.existingConfig) -}}
 apiVersion: v1
 kind: ConfigMap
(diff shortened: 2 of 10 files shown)


michaellzc:
* main
  * **PR #77** <a href="https://app.graphite.dev/github/pr/sourcegraph/deploy-sourcegraph-helm/77" target="_blank">Graphite</a>  (this PR)

charts/sourcegraph/templates/pgsql/pgsql.ConfigMap.yaml
```diff
@@ -1,3 +1,6 @@
+{{- if and .Values.pgsql.enabled (and .Values.pgsql.existingConfig .Values.pgsql.additionalConfig) -}}
+{{- fail "You can only define one of 'pgsql.existingConfig' and 'pgsql.additionalConfig' at a time" }}
+{{- end }}
```
michaellzc, line 3: I think this is valuable cuz defining both `existingConfig` and `additionalConfig` could cause confusion where only `existingConfig` will be used. It's better to fail instead of "work" sliently

charts/sourcegraph/templates/codeinsights-db/codeinsights-db.ConfigMap.yaml
```diff
@@ -1,3 +1,6 @@
+{{- if and .Values.codeInsightsDB.enabled (and .Values.codeInsightsDB.existingConfig .Values.codeInsightsDB.additionalConfig) -}}
+{{- fail "You can only define one of 'codeInsightsDB.existingConfig' and 'codeInsightsDB.additionalConfig' at a time" }}
```
caugustus-sourcegraph, line 2: Nice!

charts/sourcegraph/values.yaml
```diff
@@ -144,6 +144,10 @@ codeInsightsDB:
       value: password
   # -- Name of existing ConfigMap for `codeinsights-db`. It must contain a `postgresql.conf` key.
   existingConfig: "" # Name of an existing configmap
+  # -- Additional PostgreSQL configuration. This will overwrite our default configuration.
```
caugustus-sourcegraph, line 147: This wording is a bit unclear - it makes it sound like the entire configuration will be overwritten. Instead, we should make it obvious that this is to provide overrides or extend the existing configuration - not replacing it wholesale (like the `existingConfig` option).

charts/sourcegraph/tests/postgresAdditionalConfig_test.yaml
```diff
@@ -0,0 +1,36 @@
...
+- codeinsights-db/codeinsights-db.ConfigMap.yaml
+tests:
+- it: should have additional config rendered when *.additionalConfig is defined
+  set:
+    pgsql:
+      additionalConfig: |
+        max_connections = 1000000000000000000
+    codeIntelDB:
+      additionalConfig: |
+        max_connections = 1000000000000000000
```
caugustus-sourcegraph, line 15: I like to use different values for each test to avoid false positives - right now, the test would pass if the template were accidentally always referring to `pgsql.additionalConfig` instead of their own references.

charts/sourcegraph/values.yaml
michaellzc: I was probably thinking "override" in my brain but ended up typing "overwrite"

caugustus-sourcegraph (approved):

PR #85, opened 2022-04-01
Switch minio to use a secret for credentials, replace deprecated env vars

caugustus-sourcegraph:
Prototype to demonstrate providing an interface for minio credentials that allows us to generate or reuse an existing Secret resource instead of use direct env vars. I haven't fully tested this approach since it's still in the proof-of-concept phase.

**Decisions / consequences:**
- This implementation is opinionated about the structure of the minio-auth Secret, which simplifies the logic and allows us to easily reference the credentials in other deployments. Customers who want to use a different structure can use kustomize to alter the configuration.
- We're not randomly generating values for the credentials to preserve compatibility with the helm template command (we're stuck with either generating a new value every time helm runs, or requiring access to the existing secret).
- I went ahead and swapped the credentials to use the newer minio variables. As far as I can tell it's a naming switch only and should be backwards compatible.
- I added a checksum to force dependent pods to restart when the secret changes. This worked with a subchart but I don't know if there are other consequences to this that make it unsuitable.

Partially addresses https://github.com/sourcegraph/sourcegraph/issues/32694

### Checklist

### Test plan

- Existing secret, custom credentials
- Generated secret, custom credentials
- Upgrade from previous chart

### DIFF
--- a/charts/sourcegraph/templates/minio/minio.Secret.yaml
+++ b/charts/sourcegraph/templates/minio/minio.Secret.yaml
@@ -1,0 +1,14 @@
+{{- if and .Values.minio.enabled (not .Values.minio.auth.existingSecret) }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: minio-auth
+  labels:
+    app: minio
+    deploy: sourcegraph
+    app.kubernetes.io/component: minio
+type: Opaque
+data:
+  user: {{ .Values.minio.auth.user | toString | b64enc | quote }}
+  password: {{ .Values.minio.auth.password | toString | b64enc | quote }}
+{{- end -}}
--- a/charts/sourcegraph/templates/precise-code-intel/worker.Deployment.yaml
+++ b/charts/sourcegraph/templates/precise-code-intel/worker.Deployment.yaml
@@ -28,7 +28,7 @@ spec:
     metadata:
       annotations:
         kubectl.kubernetes.io/default-container: precise-code-intel-worker
-        checksum/config: {{ include (print $.Template.BasePath "/minio/minio.Secret.yaml") . | sha256sum }}
+        checksum/minio.secret: {{ include (print $.Template.BasePath "/minio/minio.Secret.yaml") . | sha256sum }}
       {{- if .Values.sourcegraph.podAnnotations }}
       {{- toYaml .Values.sourcegraph.podAnnotations | nindent 8 }}
       {{- end }}
--- a/charts/sourcegraph/templates/precise-code-intel/worker.Deployment.yaml
+++ b/charts/sourcegraph/templates/precise-code-intel/worker.Deployment.yaml
@@ -28,6 +28,7 @@ spec:
     metadata:
       annotations:
         kubectl.kubernetes.io/default-container: precise-code-intel-worker
+        checksum/config: {{ include (print $.Template.BasePath "/minio/minio.Secret.yaml") . | sha256sum }}
       {{- if .Values.sourcegraph.podAnnotations }}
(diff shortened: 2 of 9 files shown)


charts/sourcegraph/templates/minio/minio.Secret.yaml
```diff
@@ -0,0 +1,14 @@
...
+kind: Secret
+metadata:
+  name: minio-auth
+  labels:
+    app: minio
+    deploy: sourcegraph
+    app.kubernetes.io/component: minio
+type: Opaque
+data:
+  user: {{ .Values.minio.auth.user | toString | b64enc | quote }}
```
caugustus-sourcegraph, line 12: @michaellzc - What's your opinion on this Secret structure? Your proposal in https://github.com/sourcegraph/sourcegraph/issues/32694 was much more flexible. Would that be a better approach?

michaellzc: I like your approach, it's more opinionated.

Nit-pick: What's your thought on renaming the key from `user` and `password` to `awsAccessKeyId`  and `awsSecretAccessKey`? Using the AWS S3's naming convention feels more natural to me

charts/sourcegraph/templates/precise-code-intel/worker.Deployment.yaml
```diff
@@ -28,6 +28,7 @@ spec:
     metadata:
       annotations:
         kubectl.kubernetes.io/default-container: precise-code-intel-worker
+        checksum/config: {{ include (print $.Template.BasePath "/minio/minio.Secret.yaml") . | sha256sum }}
```
michaellzc, line 31: I would probably rename the annotation key to something more specific, such as `checksum/minio-secret`,  in case we need more checksum in the future.

michaellzc (commented):
Do you think we should create abstractions for [external object storage](https://docs.sourcegraph.com/admin/external_services/object_storage) as well? Is that out of the scope? Now I feel this implementation is kind of incomplete. It's worth taking a look at how gitlab helm chart is handling it https://docs.gitlab.com/charts/advanced/external-object-storage/

😬 I really wish there is a site-config.json equivalent to configure the sourcegraph deployment, for example

`server-config.yaml`

```yaml
objectStorage:
  s3:
    accessKeyId: ""
    secretAccessKey: ""
  // or use gcs
  gcs:
    // stuff

pgsql:
  connectionUri: ""

codeIntelDb:
  connectionUri: ""
```

I'll take a look at the database abstraction PR later and go over some examples from gitlab Maybe we need some sync time next week to go over it. I am feeling really indecisive

charts/sourcegraph/templates/precise-code-intel/worker.Deployment.yaml
caugustus-sourcegraph: 👍  I started to swap it in the db PR but didn't make the same change here, will update.

michaellzc:
> > Do you think we should create abstractions for external object storage
> 
> It seems a bit out of scope for this PR but I agree that we should do some thinking here in case it affects this implementation.
> 
> > I really wish there is a site-config.json equivalent to configure the sourcegraph deployment
> 
> Agreed, the tension is especially apparent for the DB PR: the `pgsql` deployment can be disabled, but you still need to configure the credentials there. It's a weird mix of general configuration + deployment details. I'll do some more thinking too and we can figure out a good approach here. It was useful just to open these PR's and see what rough edges came out, but I'm not feeling good about the actual implementation.

definitely! thanks a lot for putting the PoC together or we wouldn't be able to see all these rough edges :)

michaellzc (approved):

PR #86, opened 2022-04-01
Use secrets for pgsql credentials

caugustus-sourcegraph:
Similar to https://github.com/sourcegraph/deploy-sourcegraph-helm/pull/85, but for database credentials.

Prototype to demonstrate providing an interface for database credentials that allows us to generate or reuse an existing Secret resource instead of use direct env vars. This isn't a full implementation (missing codeintel / codeinsights) since it's still a proof of concept. The migrator chart would also need to be updated.

Decisions / consequences:
* Just like the minio implementation, this is opinionated about the structure for the Secret. Is this riskier with auth for an external database?
* This approach vastly simplifies the management of database credentials - they're specified in one location now, instead of 4 different places.
* The variable names are all over the place (example: PGDATABASE, POSTGRES_DB, DATA_SOURCE_URI). I couldn't think of a clean solution for handling this so it's pretty messy. Really hoping someone else has a good idea!
* I'm on the fence about whether `host` and `database` belong in the secret. Should those be normal variables instead?

Partially addresses https://github.com/sourcegraph/sourcegraph/issues/32694

### Checklist

### Test plan

Manual test plan + confirming new functionality

### DIFF
--- a/charts/sourcegraph/values.yaml
+++ b/charts/sourcegraph/values.yaml
@@ -589,7 +589,7 @@ pgsql:
   enabled: true
   auth:
     # -- Name of existing secret to use for Postgres credentials
-    # The secret must contain the keys `user`, `password`, `database` and `host`.
+    # The secret must contain the keys `user`, `password`, `database`, `host` and `port`.
     # `auth.user`, `auth.password`, etc. are ignored if this is enabled
     existingSecret: ""
     # -- Sets postgres database name
@@ -600,6 +600,8 @@ pgsql:
     user: "sg"
     # -- Sets postgres password
     password: "password"
+    # -- Sets postgres port
+    port: "5432"
   # -- Name of existing ConfigMap for `pgsql`. It must contain a `postgresql.conf` key
   existingConfig: "" # Name of an existing configmap
   # -- Additional PostgreSQL configuration. This will override or extend our default configuration.
--- a/charts/sourcegraph/values.yaml
+++ b/charts/sourcegraph/values.yaml
@@ -262,12 +262,6 @@ frontend:
       value: http://grafana:30070
     JAEGER_SERVER_URL:
       value: http://jaeger-query:16686
-    PGDATABASE:
-      value: sg
-    PGHOST:
-      value: pgsql
-    PGUSER:
-      value: sg
     PROMETHEUS_URL:
       value: http://prometheus:30090
   image:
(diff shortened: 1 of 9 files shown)


michaellzc (commented):
Looks great and I really like the abstraction!

> I'm on the fence about whether host and database belong in the secret. Should those be normal variables instead?

I think everything inside a secret is a good approach, instead of all over different places.

### Password

I think we're not enforcing password for the built-in pg before right? Would this cause any problems to existing helm customers during upgrade?

### Usage for external databases and pgbouncer

Is the usage of external databases within the scope of this PR?

I am wondering how do we achieve the same thing described here https://docs.sourcegraph.com/admin/external_services/postgres#kubernetes-1 in helm

We have to pass `statement_cache_mode=describe` to the `PGDATASOURCE` env var, and this PR doesn't provide any abstraction to customize the `PGDATASOURCE`. I think the user would have to fall back to handcrafting the `frontend.env` values on their own.

charts/sourcegraph/values.yaml
```diff
@@ -564,6 +558,19 @@ minio:
 pgsql:
   # -- Enable `pgsql` PostgreSQL server
   enabled: true
```
caugustus-sourcegraph, line 560: @michaellzc - After giving this abstraction some more thought, I actually think the implementation as-is will be a sufficient improvement that we should keep it (and the related minio abstraction).

I'm happy enough with the credentials being stored in their current location, but I'm also ok if we want to move them to a top-level setup, e.g.:

`values.yaml`:
```
credentials:
  pgsql:
    existingSecret:
    database: "sg"
    ...
  minio:
    user: "asdfadsf"
    ...
```

If we keep the credentials alongside the deployments, we could also consider renaming the `pgsql.enabled` flag to `pgsql.external` to make it more clear what that flag will do. I'm uneasy about this though, seems like `enabled` is more standard.

Arguments against the current abstraction:
- **It's incomplete:** if customers want to use an external object store, or use pgbouncer, they'll need to set some environment variables by hand. I don't see this as a significant limitation - they already need to do this, and adding the abstraction doesn't hinder their ability to do this.
- **We may modify this abstraction later and require migration:** We actually reduce the disruption required by enforcing a standard now, because credentials will be stored in a predictable location / single source of truth. If we launch without any abstraction, the database credentials are repeated several times throughout the values.yaml and migration will be worse.

I think the abstraction helps more than it hurts, even if we end up changing something later. It simplifies the process of customizing database/minio credentials, promotes a better practice (using Secrets), and gives us a standard set of keys that we can reference for future migrations.

michaellzc: I am sold. let's go with the abstraction.

michaellzc (approved):

PR #87, opened 2022-04-04
Fix ingress settings on GCP example

caugustus-sourcegraph:
The example given for GCP did not actually create a load balancer. According to [the docs](https://cloud.google.com/kubernetes-engine/docs/how-to/custom-ingress-controller#controller_summary), GKE relies on the `kubernetes.io/ingress.class` annotation. If that annotation is not set but ingressClassName is, no load balancer is provisioned.

Supposedly you're able to specify both `kubernetes.io/ingress.class` and `ingressClassName`, but in practice I got an error when trying to provide both, so I removed it.

I'll update the example in the main docs as well, but there's an open PR + a couple other changes we might want to make before then.

### Checklist

- [X] Follow the [manual testing process](https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/main/TEST.md)

### Test plan

Install the chart, confirm a load balancer gets created.

### DIFF
--- a/charts/sourcegraph/examples/gcp/override.yaml
+++ b/charts/sourcegraph/examples/gcp/override.yaml
@@ -3,8 +3,7 @@ frontend:
   ingress:
     enabled: true
     annotations:
-      kubernetes.io/ingress.class: null
-    ingressClassName: gce
+      kubernetes.io/ingress.class: gce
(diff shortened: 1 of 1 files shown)


michaellzc (approved):
thank you!

PR #88, opened 2022-04-04
Prototype for including extra manifests in the helm values file

caugustus-sourcegraph:
Simple proof-of-concept for allowing additional manifests to be included in the values file. I can't decide if this is a good idea!

Pros:
- Simple manifests can be embedded in the values file, avoiding the need to use kustomize or subcharts
- We can remove the manual step to deploy a BackendConfig from the gce demo
- Can reference other values since we're templating

Cons:
- Dangerous, very easy to mess up

### Checklist

### DIFF
--- a/charts/sourcegraph/examples/gcp/override.yaml
+++ b/charts/sourcegraph/examples/gcp/override.yaml
@@ -17,3 +17,26 @@ storageClass:
   provisioner: pd.csi.storage.gke.io
   volumeBindingMode: WaitForFirstConsumer
   reclaimPolicy: Retain
+
+extraManifests: |
+  apiVersion: cloud.google.com/v1
+  kind: BackendConfig
+  metadata:
+    name: sourcegraph-frontend
+  spec:
+    healthCheck:
+      checkIntervalSec: 5
+      timeoutSec: 5
+      requestPath: /ready
+      port: 6060 # we use a custom port to perform healthcheck
+  ---
+  apiVersion: cloud.google.com/v1 # Second manifest just for demonstration purposes
+  kind: BackendConfig
+  metadata:
+    name: sourcegraph-frontend-example
+  spec:
+    healthCheck:
+      checkIntervalSec: 5
+      timeoutSec: 5
+      requestPath: /ready
+      port: 6060 # we use a custom port to perform healthcheck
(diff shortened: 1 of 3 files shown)


michaellzc (approved):
This is great.

Maybe we can reuse the `raw` chart? (either make it a depenency or just copy the template)
https://github.com/helm/charts/blob/master/incubator/raw/templates/resources.yaml

It feels a bit unnatural to write YAML as a string inside a YAML.

it's easy enough to have a typo or broken indentation in YAML already, this feels worse? Like IDE usually warns us when we make obvious mistakes, but if we make it a string, user won't be able to benefit from it. Also, in the future we may be able to publish the json schema of our values file and maybe we can get autocomplete k8s manifests working for `additionalManifests` in VS Code.

https://github.com/helm/charts/tree/master/incubator/raw

caugustus-sourcegraph:
@michaellzc - Ah, that raw chart is great! I didn't know about that.

I updated the implementation to try and follow that chart's pattern a bit closer - is this more like what you were thinking?

michaellzc:
sorry, I should have been more specific

What's your thought on 

```yaml
extraResources:
- |
  apiVersion: cloud.google.com/v1
  kind: BackendConfig
  metadata:
    name: sourcegraph-frontend
  spec:
    healthCheck:
      checkIntervalSec: 5
      timeoutSec: 5
      requestPath: /ready
      port: 6060 # we use a custom port to perform healthcheck
```

vs 

```yaml
extraResources:
- apiVersion: cloud.google.com/v1
  kind: BackendConfig
  metadata:
    name: sourcegraph-frontend
  spec:
    healthCheck:
      checkIntervalSec: 5
      timeoutSec: 5
      requestPath: /ready
      port: 6060 # we use a custom port to perform healthcheck
```

I prefer the latter. It may address one of your concern `Dangerous, very easy to mess up`.

- IDE can help make sure it's not broken YAML. It's just a string right now, I think it's less likely to mess up?
- we can publish jsonschema in the future where IDE that supports such feature can treat object under the `extraResources` as kubenrnetes object and provide auto complete

thoughts?

michaellzc:
> @michaellzc - Is there a way to get your preferred option and support templating at the same time?

this is some dark 🪄

partial template -> https://github.com/bitnami/charts/blob/master/bitnami/common/templates/_tplvalues.tpl

template -> https://github.com/bitnami/charts/blob/master/bitnami/mongodb/templates/extra-list.yaml

caugustus-sourcegraph:
@michaellzc 👍  OK! How about now? (Thank you for your example charts, they've been super useful)

michaellzc (approved):

PR #89, opened 2022-04-05
Update GKE example to remove manual steps

caugustus-sourcegraph:
Updates the GKE example so it works out-of-the-box without any manual steps. Adding a domain / TLS is still an additional step but isn't required for this example.

### Checklist

### Test plan

New deployment using the provided override file succeeded.


michaellzc (approved):
yay

PR #94, opened 2022-04-06
Add mapping between values key and names used for multi-word resources

caugustus-sourcegraph:
**Goal**: I need to be able to reference the auto-generated database secret by name. The secret name is hardcoded (which is not ideal) but even if it were generated, the same problem would occur (example: we would still need to be able to predict the `"codeintel-db"` section of [this line](https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/main/charts/sourcegraph/templates/codeintel-db/codeintel-db.StatefulSet.yaml#L14)).

My untested hack to use `snakecase $service` to convert `codeIntelDB` to `codeintel-db` didn't work (it produces `code-intel-db`, of course).

This approach is pretty much me admitting defeat: I can't think of a better way to map codeIntelDB to codeintel-db. It's a little goofy putting this mapping into the values.yaml since it's not intended to be customized, but maybe leaving it in un-documented will be sufficient?

### Checklist

### Test plan

A followup PR actually generates the correct secret references

### DIFF
--- a/charts/sourcegraph/templates/_helpers.tpl
+++ b/charts/sourcegraph/templates/_helpers.tpl
@@ -129,11 +129,21 @@ Jaeger selector labels
 app.kubernetes.io/name: jaeger
 {{- end }}
 
+{{- define "sourcegraph.mappedName" -}}
+{{- $top := index . 0 -}}
+{{- $service := index . 1 -}}
+{{- if (index $top.Values $service "mappingName") }}
+{{- (index $top.Values $service "mappingName") }}
+{{- else }}
+{{- $service }}
+{{- end -}}
+{{- end }}
+
 {{- define "sourcegraph.databaseAuth" -}}
 {{- $top := index . 0 -}}
 {{- $service := index . 1 -}}
 {{- $prefix := index . 2 -}}
-{{- $secretName := ((snakecase $service) | replace "_" "-") -}}
+{{- $secretName := (include "sourcegraph.mappedName" (list $top $service )) }}
 {{- $secretName := printf "%s-auth" $secretName -}}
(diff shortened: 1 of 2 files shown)


michaellzc:
I ran into a similar problem in https://github.com/sourcegraph/deploy-sourcegraph-helm/pull/52#discussion_r816312728

https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/1a6d3ab171d39e30f6cd9d73aa45be223a677d68/charts/sourcegraph/templates/_helpers.tpl#L41-L46

What's your thought on introducing another breaking change? e.g., we rename the values `codeInsightDB` -> `codeintelDb`?

If not, I think we should update `serviceAccoutName` partial template to follow the same conventino.

caugustus-sourcegraph:
@michaellzc Actually I realized I'm dumb for introducing a new variable - couldn't we just prepopulate `codeIntelDB.name` with `codeintel-db`, instead of adding `codeIntelDB.mappingName`? And then your `serviceAccountName` logic could be simpler too (` {{- $defaultServiceAccountName := (index $top.Values $service "name")`)

michaellzc:
> @michaellzc Actually I realized I'm dumb for introducing a new variable - couldn't we just prepopulate `codeIntelDB.name` with `codeintel-db`, instead of adding `codeIntelDB.mappingName`? And then your `serviceAccountName` logic could be simpler too:
> 
> ```
>  {{- $defaultServiceAccountName := (index $top.Values $service "name")
>  {{- default $defaultServiceAccountName (index $top.Values $service "serviceAccount" "name") }} 
> ```

🙃 I was dumb too.

Yes!

caugustus-sourcegraph:
Great work everyone, we discovered this is a bad idea

PR #97, opened 2022-04-06
Set default name for resources

caugustus-sourcegraph:
Alternate implementation of https://github.com/sourcegraph/deploy-sourcegraph-helm/pull/94 - sets a default name for all services so that we can easily reference resources.

The second commit updates most resources to use this default name, instead of a hardcoded version. I'm on the fence about how valuable this is.

**Not included**: service names and persistentVolumeClaim names remain hardcoded, instead of referencing the variable. I'm unclear what impact changing a service name would have on Sourcegraph's service discovery (most likely we can work around it by setting environment variables, but it's easier to avoid creating this problem). For PVC's, it seems risky to change the name if it had previously been created.

### Checklist

### Test plan

No change to rendered output before/after using default values.


michaellzc:
> service names and persistentVolumeClaim names remain hardcoded, instead of referencing the variable. I'm unclear what impact changing a service name would have on Sourcegraph's service discovery (most likely we can work around it by setting environment variables,

I can confirm changing service names will break service discovery. We have way too many magic default env var e.g. `GITSERV_URL` (or something like that), and it's not fun to haunt each one down.

👍 let's just don't touch it.

changing PVC name should be fine, but there's no good reason to do it, so let's leave it.

michaellzc (approved):

PR #100, opened 2022-04-07
Add abstraction for redis connections

caugustus-sourcegraph:
Creates an abstraction for the redis endpoints, so that using an external redis is easier. Examples/docs ~~not yet updated and we need tests, will update if this gets approved.~~ are now updated.

I also changed the checksum annotations to be aggregates, instead of having a bunch of individual annotations.

**Questions**:
- I don't love the `redisStore.connection.endpoint` `.existingSecret` setup. Would this be better as a top-level `redisStore.endpoint` and `redisStore.existingSecret`? I keep thinking 'maybe we'll add password support' and that would make more sense as another key underneath `connection`.
- This approach forces the use of two secrets instead of one (like the example in https://github.com/sourcegraph/deploy-sourcegraph-helm/tree/main/charts/sourcegraph/examples/external-redis#redis_endpoint). Too annoying?
- If there's an easier way to get the aggregate checksums, please share! This is the only thing I could come up with but I'm probably missing something.

### Checklist


michaellzc:
> This approach forces the use of two secrets instead of one (like the example in [https://github.com/sourcegraph/deploy-sourcegraph-helm/tree/main/charts/sourcegraph/examples/external-redis#redis_endpoint](https://github.com/sourcegraph/deploy-sourcegraph-helm/tree/main/charts/sourcegraph/examples/external-redis?rgh-link-date=2022-04-07T21%3A43%3A34Z#redis_endpoint)). Too annoying?

I think it makes more sense to have two. I feel like it kind of emphasizes the fact you REALLY should use two separate redis instances externally.

we should update the example to reflect that.

> I don't love the redisStore.connection.endpoint .existingSecret setup. Would this be better as a top-level redisStore.endpoint and redisStore.existingSecret? I keep thinking 'maybe we'll add password support' and that would make more sense as another key underneath connection.

I don't have a strong preference. A use case I could think of, an external redis instance that uses self-signed cert.

```
redisStore:
  connection:
    endpoint: redis://stuff
    caCert: |
      # ... some cert
```

I don't think sourcegraph supports configuring a custom ca certificate unless the customer forks our image and bake it into the trust store

> If there's an easier way to get the aggregate checksums, please share! This is the only thing I could come up with but I'm probably missing something.

😬 not that I know of

but maybe this?

```tpl
checksum/haha: {{ toJson .Values | sha256sum }}
```

michaellzc (approved):

caugustus-sourcegraph:
@michaellzc - ```checksum/haha: {{ toJson .Values | sha256sum }}``` was a fantastic suggestion, thank you! The implementation is much simpler now, and using that removed an annoying problem in the tests.

This approach should work even if we start to include a caCert in the connection info.

caugustus-sourcegraph:
Re-requesting review for the new tests / example changes - the tests are pretty fragile but they work well enough, and we're unlikely to really adjust those defaults.

charts/sourcegraph/tests/checksum_test.yaml
```diff
@@ -0,0 +1,60 @@
...
+tests:
+- it: should update checksum when pgsql secret is modified
+  set:
+    pgsql:
+      auth:
+        existingSecret: "my-secret"
+  asserts:
+      - equal:
+          path: spec.template.metadata.annotations.[checksum/auth]
+          value: "b960ae6db8c72ca9248278a63593de67f1ddfb9805f82eded969ffde767406db"
```
michaellzc, line 14: I think this would be a good use case of https://github.com/quintush/helm-unittest/blob/master/README.md#snapshot-testing

michaellzc (approved):
not blocking

charts/sourcegraph/tests/checksum_test.yaml
caugustus-sourcegraph: Thanks, updated. Snapshots don't make sense for the initial feature development (I need to actually establish that the annotation changes first), but it does help now.

charts/sourcegraph/examples/external-redis/override.yaml
```diff
@@ -0,0 +1,12 @@
...
+
+redisCache:
+  enabled: false
+  connection:
+    endpoint: redis://redis-cache.example.com:6379/2
+
+redisStore:
+  enabled: false
+  connection:
+    endpoint: redis://redis-store.example.com:6379
```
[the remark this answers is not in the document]
caugustus-sourcegraph: Added some clarification - does that help?

PR #103, opened 2022-04-08
Document hotfix process

caugustus-sourcegraph:
Document the steps to take in the (hopefully) unusual situation that we need to create a hotfix for the helm chart itself.

Not sure if this should move into the handbook - maybe we can leave it here until we use it once, and then migrate?

### Checklist

### Test plan

None, doc only


michaellzc (approved):
I really hope we won't have to do any hotfix

PR #108, opened 2022-04-18
fix: workaround for r/o rootfs support for prometheus, gitserver

kevinwojo:
### Checklist

### Test plan

Follow the manual testing process, also validating on dogfood: https://github.com/sourcegraph/deploy-sourcegraph-dogfood-k8s/pull/4203


charts/sourcegraph/values.yaml
```diff
@@ -401,7 +401,9 @@ gitserver:
     allowPrivilegeEscalation: false
     runAsUser: 100
     runAsGroup: 101
-    readOnlyRootFilesystem: true
+    # -- Read-only filesystem not supported for the gitserver container,
+    # see [sourcegraph/issues/34012](https://github.com/sourcegraph/sourcegraph/issues/34012) for more information
+    readOnlyRootFilesystem: false
```
caugustus-sourcegraph, line 406: For gitserver - maybe instead of disabling readOnly, we can try adding an emptydir `/tmp` volume mount? Similar to [searcher here](https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/ace2ed4f57202b90d4b01c84e4f2d0ffba930587/charts/sourcegraph/templates/searcher/searcher.Deployment.yaml#L92-L93).

charts/sourcegraph/values.yaml
```diff
@@ -729,7 +731,9 @@ prometheus:
     allowPrivilegeEscalation: false
     runAsUser: 100
     runAsGroup: 100
-    readOnlyRootFilesystem: true
+    # -- Read-only filesystem not supported for the prometheus container,
```
caugustus-sourcegraph, line 734:
```suggestion
    # Read-only filesystem not supported for the prometheus container,
```
I don't think these need to generate an individual entry in the README

charts/sourcegraph/values.yaml
kevinwojo: Tested this out, looks good on my end.

caugustus-sourcegraph (approved):

PR #104, opened 2022-04-09
Add insiders release

michaellzc:
This is currently live at https://github.com/sourcegraph/deploy-sourcegraph-helm-fork/tree/gh-pages/insiders

try it out

```sh
helm repo add sg-fork-nightly https://sourcegraph.github.io/deploy-sourcegraph-helm-fork/insiders
helm search repo -l sg-fork --devel
```

## Notes

Ideally, we want to publish insiders build in a different git repo, but it's currently not possible with `chart-releaser`. Its implementation does not respect the configured `git-repo` and `owner` when interacting with `index.yaml` in `cr index`, these two flags are only used during the `cr upload` process.

Also, prior to merging this PR, we need to manually upload the initial `index.yaml`. I think something must change recently and caused such weird behaviour. Follow the step below

```sh
cr package charts/sourcegraph --config insiders-config.yaml
cr package charts/sourcegraph-migrator --config insiders-config.yaml
export CR_TOKEN="<GH_TOKEN>" # https://start.1password.com/open/i?a=HEDEDSLHPBFGRBTKAKJWE23XX4&v=dnrhbauihkhjs5ag6vszsme45a&i=7s46bqcnl5hxzbutupu44va7gu&h=team-sourcegraph.1password.com
cr upload --config insiders-config.yaml
mkdir -p insiders
cr index --config insiders-config.yaml
```

Manually upload the `insiders/index.yaml` to the `gh-pages` branch.

## More notes

I really don't like this approach and feel really hacky to me. Also, the release page is not pretty https://github.com/sourcegraph/deploy-sourcegraph-helm-fork/releases

An alternative would be moving the chart repo to a GCS bucket, and we can just copy and paste whatever linked is doing

- https://github.com/linkerd/linkerd2/blob/main/.github/workflows/release.yml#L353
- https://github.com/linkerd/linkerd2/blob/main/.github/actions/helm-publish/action.yml

### Checklist

- [ ] Follow the [manual testing process](https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/main/TEST.md)

### Test plan

see pr description

PR #112, opened 2022-04-21
Set up publishing to GCS

caugustus-sourcegraph:
Modification of https://github.com/sourcegraph/deploy-sourcegraph-helm/pull/104, to publish to GCS instead of github pages.

This configuration has not been tested (blocked waiting for some other changes), so there may be quite a few changes required. However, I'd like to get feedback on the overall structure/approach in the meantime.

In particular:
- Having separate workflow files for insiders vs releases. They're mostly identical, but it seems better to keep them separate instead of adding complicated conditional logic. Any concerns?
- Charts will be published to helm.sourcegraph.com/insiders and helm.sourcegraph.com/release. These will be used during the `helm repo add` commands. The names feel awkward to me - suggestions? `edge` and `stable` don't seem bad (copying from the linkerd2 setup @michaellzc mentioned before).
- Same as #104, on insiders builds every chart will be published even if there's no change. Maybe it's a problem we defer and fix later, if at all?
 
### Checklist

### Test plan

Confirm releases and main commits get published to the GCS bucket.


michaellzc:
> Having separate workflow files for insiders vs releases. They're mostly identical, but it seems better to keep them separate instead of adding complicated conditional logic. Any concerns?
LGTM

> Charts will be published to helm.sourcegraph.com/insiders and helm.sourcegraph.com/release. These will be used during the helm repo add commands. The names feel awkward to me - suggestions?

`insiders` sounds like a type of `release`, and we use `release` to reference the stable release. However, it's worth noting that we don't really have the concept of a "stable" release or multiple release channels in Sourcegraph, it's just a monthly branch cut. So I think we can just keep it the way it is right now.

> Same as https://github.com/sourcegraph/deploy-sourcegraph-helm/pull/104, on insiders builds every chart will be published even if there's no change. Maybe it's a problem we defer and fix later, if at all?

Yeah, I don't think it's worth putting too much thought into it right now.

michaellzc (approved):
lgtm

I like the idea of using https://github.com/hayorov/helm-gcs so we don't have to handle the "sync" logic on our own using gsutil.

PR #113, opened 2022-04-22
Set public url when pushing to GCS

caugustus-sourcegraph:
By default, helm-gcs uses `gs://<bucket name>` as the public URL for the index.yaml, and downloading the chart requires installing the helm-gcs plugin.

To avoid adding this dependency (and actually use Cloudflare), configure the https public url during pushes.

### Checklist

### Test plan

helm chart install succeeds without the helm-gcs plugin.


kevinwojo (approved):
SGTM

PR #114, opened 2022-04-22
Merge release into main

caugustus-sourcegraph:
When we created the release/3.39 branch we updated the chart versions. This needs to be merged back into main so that future releases are also updated.

We'll need to modify the release process to include this step automatically - maybe by creating another PR?

### Checklist

### Test plan

Chart publishes to insiders repo with the 3.39 version number.


kevinwojo (approved):
Also noting that we've implicitly agreed to tie the Chart version and appVersion to Sourcegraph release version.

PR #115, opened 2022-04-22
Add quotes to chart version replacement

caugustus-sourcegraph:
The release batch change added quotes around the chart version, so the sed used for insiders releases no longer works.

Probably this could be done in a more elegant / graceful way but adding the quotes is pretty easy to unblock us for now.

### Checklist

### Test plan

Chart publish should work this time


kevinwojo (approved):
That's a wrap!

PR #120, opened 2022-05-04
Fixed creation frontend rolebinding when privileged is false

Lawliet2012:
### Checklist

### Test plan

I have K8s cluster which I'm using as a client without some permissions like "Creating RoleBindings". I found that prometheus, when you using 
```
prometheus:
    privileged: false
```
doesn't create roleBindings and it's working perfectly. So I did same for frontend part of chart and everything works perfectly on my own dev cluster and deployed on staging, so I can use sourcegraph without any troubles with my changes.

### DIFF
--- a/charts/sourcegraph/templates/frontend/sourcegraph-frontend.RoleBinding.yaml
+++ b/charts/sourcegraph/templates/frontend/sourcegraph-frontend.RoleBinding.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.frontend.privileged -}}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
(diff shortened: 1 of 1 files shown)


michaellzc:
+1 on Crystal's points, this deserves a more well-thought solution, a standalone `rbac` chart to deploy privileged resources.

I would also like to point out we have a tutorial to demonstrate how to use kustomize as an escape hatch when there are things Sourcegraph doesn't support at the moment.

You may check it out here https://docs.sourcegraph.com/admin/deploy/kubernetes/helm#integrate-kustomize-with-helm-chart.

Lawliet2012:
> In the interest of unblocking your change - what do you think about introducing a new flag that specifically toggles rbac creation for the frontend, separate from the privileged setting, and defaulting it to true?

Sorry for a bit delay. Yes, this is exactly what I want, so I can rename my own flag to something like `.Values.frontend.rbacCreation` and make it default in `values.yml`

beyang:
Closing for staleness. Please re-open if still relevant.

PR #124, opened 2022-05-17
Update alpine-3.12 references

caugustus-sourcegraph:
The alpine-3.12 image was removed in https://github.com/sourcegraph/sourcegraph/pull/34906 (as a followup to https://github.com/sourcegraph/sourcegraph/pull/34508). We directly reference the alpine-3.12 image here to run init commands, so the reference needs to be updated.

Note: the insiders reference will be updated as part of the next release.

### Checklist

### Test plan

Not tested directly apart from verifying the image reference - however, the same change has already been made in https://github.com/sourcegraph/deploy-sourcegraph-dogfood-k8s/pull/4336 and has been working fine.

### DIFF
--- a/charts/sourcegraph/CHANGELOG.md
+++ b/charts/sourcegraph/CHANGELOG.md
@@ -9,6 +9,7 @@ All notable changes to Sourcegraph are documented in this file.
 ### Changed
 
 - **IMPORTANT** `cadvisor` now defaults to run in `privileged` mode. This allows `cadvisor` to collect out of memory events happening to containers which can be used to discover underprovisoned resources. If you have your own monitoring infrastructure, you may choose to disable `cadvisor` or set `cadvisor.containerSecurityContext.privileged=false` in your override file. [#121](https://github.com/sourcegraph/deploy-sourcegraph-helm/pull/121)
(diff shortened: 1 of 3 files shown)


charts/sourcegraph/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ All notable changes to Sourcegraph are documented in this file.
 ### Changed
 
 - **IMPORTANT** `cadvisor` now defaults to run in `privileged` mode. This allows `cadvisor` to collect out of memory events happening to containers which can be used to discover underprovisoned resources. If you have your own monitoring infrastructure, you may choose to disable `cadvisor` or set `cadvisor.containerSecurityContext.privileged=false` in your override file. [#121](https://github.com/sourcegraph/deploy-sourcegraph-helm/pull/121)
+- The alpine-3.12 docker image has been updated to use alpine-3.14 [#124](https://github.com/sourcegraph/deploy-sourcegraph-helm/pull/124/)
```
keegancsmith, line 12: I found this a little confusing, making me think we had a `alpine-3.12` docker image, but in fact it was using alpine 3.14. Does this suggestion make sense?

```suggestion
- Uses of alpine-3.12 docker image have been updated to use alpine-3.14 [#124](https://github.com/sourcegraph/deploy-sourcegraph-helm/pull/124/)
```

keegancsmith (approved):
I see references for insiders in the changes. Is that intentional? Approving to unblock.

PR #132, opened 2022-06-10
Give ctags a writable temp dir

chrismwendt:
Because the `symbols` container's processes run as user 100 (`sourcegraph`):

https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/b6d554e3f594b9c43951f84520f174fae29114ab/charts/sourcegraph/values.yaml#L997

ctags doesn't have write permission to `/tmp`.

### Checklist

### Test plan

Tested in https://github.com/sourcegraph/deploy-sourcegraph-dogfood-k8s/pull/4594

### DIFF
--- a/charts/sourcegraph/templates/symbols/symbols.Deployment.yaml
+++ b/charts/sourcegraph/templates/symbols/symbols.Deployment.yaml
@@ -66,7 +66,7 @@ spec:
               fieldPath: metadata.name
         - name: CACHE_DIR
           value: /mnt/cache/$(POD_NAME)
-        - name: TMP_DIR
+        - name: TMPDIR
           value: /mnt/cache/$(POD_NAME)-tmp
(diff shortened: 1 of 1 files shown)


michaellzc (approved):
lgtm

but we should ask delivery to take another look

caugustus-sourcegraph (approved):

kevinwojo:
@chrismwendt do you know about how much data is to be written to this location? What will disk utilization respond to?

chrismwendt:
@kevinwojo ctags appears to write a few MB to `TMPDIR` then immediately deletes the data, so to be safe I'd say it needs 1GB of space. It reads/writes a lot, so an SSD is best.

chrismwendt:
Reviewers: I added a volume to ensure that the `/mnt/tmp` directory always exists, otherwise ctags fails.

caugustus-sourcegraph (approved):
Approving again, for the extra volume mount approach

PR #135, opened 2022-06-15
Fix storage class `type` parameter

michaellzc:
### Checklist

### Test plan

New release

### DIFF
--- a/charts/sourcegraph/Chart.yaml
+++ b/charts/sourcegraph/Chart.yaml
@@ -5,7 +5,7 @@ description: Chart for installing Sourcegraph
 type: application
 
 # Chart version, separate from Sourcegraph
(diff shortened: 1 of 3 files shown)


caugustus-sourcegraph (approved):
Maybe add a changelog entry? This is our first patch release so we don't have a great process here.

michaellzc:
> Maybe add a changelog entry? This is our first patch release so we don't have a great process here.

😂 I am having a hard time deciding where changelog update should go

- just `main`?
- or both `main` and `release/*`

also, the changelog on release branch looks out of date https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/release/3.40/charts/sourcegraph/CHANGELOG.md

PR #163, opened 2022-08-22
Don't include `annotations:` key when there are no annotations

nicksnyder:
A [customer reported lint warnings](https://sourcegraph.slack.com/archives/C03SBTYH3S9/p1661186892313959?thread_ts=1659985666.766349&cid=C03SBTYH3S9) in the output yaml from kube yaml plugin for IntellilJ because `annotation:` keys were set with no value.

This PR updates the chart to not emit the annotations key if no values are going to be set.

I am not sure how to enforce this in unit tests. `helm unittest` seems to only be capable of checking values, not the existence of keys, so tests can't tell the difference between the `annotations:` key not being present or simply empty.

Alternatively, we could just require that every entity has a description annotation.

### Checklist

### Test plan

Ran `helm template ./charts/sourcegraph` and ensured all `annotations:` keys had values.

### DIFF
--- a/charts/sourcegraph-migrator/templates/migrator/sourcegraph-migrator.Job.yaml
+++ b/charts/sourcegraph-migrator/templates/migrator/sourcegraph-migrator.Job.yaml
@@ -15,16 +15,14 @@ spec:
   backoffLimit: 5
   template:
     metadata:
-      {{- if or .Values.sourcegraph.podAnnotations .Values.migrator.podAnnotations }}
       annotations:
-        {{- if .Values.sourcegraph.podAnnotations }}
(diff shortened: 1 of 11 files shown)


michaellzc (commented):
lgtm

if you think unit test is necessary, we can consider using `matchSnapshot` https://github.com/vbehar/helm3-unittest/blob/master/DOCUMENT.md#assertion-types

CI somehow is not happy with this PR, would you take a look?

nicksnyder:
@michaellzc are you able to give me an official approve so I can merge or is there someone else I should request?

michaellzc (approved):

PR #164, opened 2022-08-22
ci: fix helm test

michaellzc:
When no change is made to the chart, the CI will fail because it tried to install the chart without setting up the kind cluster

this change ensures these tests are only run when helm chart is changed to avoid running unnecessary test

### Checklist

### Test plan

ci should pass


daxmc99 (approved):
Looks like I missed this in the first implementation! Thanks

PR #165, opened 2022-08-22
docs: set kubernetes.io/ingress.class to null if ingressClassName is set

nicksnyder:
A customer reported issues when they tried to set `ingressClassName`:

```
Error: INSTALLATION FAILED: Ingress.extensions "sourcegraph-frontend" is invalid: annotations.kubernetes.io/ingress.class: Invalid value: "nginx": can not be set when the class field is also set
```

The solution is to set `annotations.kubernetes.io/ingress.class` to `null`.

This PR updates the documentation to make this more clear.

### Checklist

### Test plan

This is just updating the documentation.


michaellzc (approved):
lgtm

would you also help fixing this example? https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/main/charts/sourcegraph/examples/ingress/override.yaml

it configures annotation and `ingressClassName` at the same time

nicksnyder:
Thanks for flagging that example! Updating here: https://github.com/sourcegraph/deploy-sourcegraph-helm/pull/166

PR #173, opened 2022-09-02
keep things dry

michaellzc:
keep things DRY

### DIFF
--- a/charts/sourcegraph/templates/indexed-search/indexed-search.StatefulSet.yaml
+++ b/charts/sourcegraph/templates/indexed-search/indexed-search.StatefulSet.yaml
@@ -55,14 +55,8 @@ spec:
         - name: {{ $name }}
           {{- $item | toYaml | nindent 10 }}
         {{- end }}
+        {{- include "sourcegraph.otelCollectorEnv" . | nindent 8 }}
         {{- if .Values.otelCollector.enabled }}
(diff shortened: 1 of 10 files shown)


charts/sourcegraph/templates/indexed-search/indexed-search.StatefulSet.yaml
```diff
@@ -55,14 +55,8 @@ spec:
...
         {{- if .Values.otelCollector.enabled }}
-        # OTEL_AGENT_HOST must be defined before OTEL_EXPORTER_OTLP_ENDPOINT to substitute the node IP on which the DaemonSet pod instance runs in the latter variable
-        - name: OTEL_AGENT_HOST
-          valueFrom:
-            fieldRef:
-              fieldPath: status.hostIP
-        - name: OTEL_EXPORTER_OTLP_ENDPOINT
-          value: http://$(OTEL_AGENT_HOST):4317
         - name: OPENTELEMETRY_DISABLED
           value: "false"
```
michaellzc, line 61: I guess zoekt has otel disabled by default?

bobheadxi: disabled = false => enabled :)

michaellzc: lol I just noticed it's `DISABLED` instead of `ENABLED`

sanderginn (changes requested):
Neat, thank you 🙂 
The source branch was just updated to name the values `openTelemetry` instead of `otelCollector`, so just need to reflect that here

sanderginn (approved):

PR #194, opened 2022-10-25
add node exporter daemonset

ggilmore:
### Overview 

_See https://github.com/sourcegraph/sourcegraph/pull/43107 for more context._

This PR adds [node-exporter](https://github.com/prometheus/node_exporter) to our kubernetes deployment.  Node-exporter is a Prometheus exporter that exposes machine-level metrics (as opposed to [cadvisor](https://github.com/google/cadvisor), which exposes _container_-level metrics). Node-exporter can expose metrics like [disk i/o stats](https://github.com/sourcegraph/sourcegraph/pull/43070), which can help us scale Sourcegraph deployments for our larger customers. 

---

Overall Node-exporter's deployement is quite similar to our existing [cadvisor](https://github.com/sourcegraph/deploy-sourcegraph-helm/tree/main/charts/sourcegraph/templates/cadvisor) configuraion:
- It needs to **run as a daemonset** (where each node-exporter instance is scraping a different node in the cluster)
- In order to gather stats from the node, node-exporter needs to mount special directories from the host such as `/`, `/proc`, and `/sys` (as readonly):
	https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/5607d33260534d5e13ce73dcd9370b5a6ea05fc8/charts/sourcegraph/templates/node-exporter/node-exporter.DaemonSet.yaml#L128-L137 
- In order to gather some metrics from the host, it needs to share the host's PID and Network namespaces:
	- https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/5607d33260534d5e13ce73dcd9370b5a6ea05fc8/charts/sourcegraph/templates/node-exporter/node-exporter.DaemonSet.yaml#L126-L127

However, unlike cadvisor, node exporter:
- Does **not need to run as root**: 
	https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/5607d33260534d5e13ce73dcd9370b5a6ea05fc8/charts/sourcegraph/values.yaml#L702-L713 
- Does **not** need to read the kernel logs (/`dev/kmsg`)
- Does **not** need to mount the host's `/var/lib/docker`
-  Uses a do-nothing services so that Prometheus can scrape all of its endpoints: https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/node-exporter/charts/sourcegraph/templates/node-exporter/node-exporter.Service.yaml

---

Node-exporter is enabled by default since it'll help us collect crucial machine-level statistics. 

However, I realize that there are users that don't want to run a daemonset of any sort / are unwilling to run services that mount `/`, `/proc`, and `/sys`. For those users, they can disable node exporter (and all of its associated resources) by setting `nodeExporter.enabled: false` in their override file.  This is documented in the changelog. 

### Other miscellaneous changes

- I tweaked the default prometheus configuration to add a new label, "nodename" to all of our metrics. This allows us to know which node in the cluster a given metric is coming from. 

### Checklist

- [ ] figure out what to do with image tag

#### unit tests

 I wrote some unit tests here: https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/node-exporter/charts/sourcegraph/tests/nodeExporter_test.yaml

 I also some wrote some "rote" unit tests that check to see if simple substitutions (such as specifying container resources) were working properly. I ended up removing these since I wasn't sure if they were high-value tests (and no other unit tests were testing these kinds of things). I can add some of them back if you think they're worth it, however). 

<details>
<summary>Example of rote test</summary>

```yaml
 - it: should propagate any set resources 
    set:
      nodeExporter:
        resources:
          limits:
            cpu: '9999'
          requests:
            memory: '1Ki'
    asserts:
      - equal:
          path: spec.template.spec.containers[0].resources.limits.cpu
          value: '9999'
      - equal:
          path: spec.template.spec.containers[0].resources.requests.memory
          value: '1Ki'
    templates: 
      - node-exporter/node-exporter.DaemonSet.yaml
```

</details>

#### local smoke test 

- I also smoke-tested this change by deploying this helm chart locally (using the instructions in [TEST.md](https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/node-exporter/TEST.md) with the following override file):

```yaml
nodeExporter:
  enabled: true
  image:
    defaultTag: 179720_2022-10-25_4d925e87cfb8@sha256:2d9dcdf0b2226f0c3d550a64d2667710265462350a3ba9ebe37d0302bc64af0f
  # extraArgs: [
  #   "--help",
  #   ]
# Disable SC creation
storageClass:
  create: false
  name: standard

sourcegraph:
  localDevMode: true
```

I then port-forwarded the grafana service and verified that I was able to pull up a metric that node-exporter provides:

image

### DIFF
--- a/charts/sourcegraph/templates/prometheus/prometheus.ConfigMap.yaml
+++ b/charts/sourcegraph/templates/prometheus/prometheus.ConfigMap.yaml
@@ -171,6 +171,11 @@ data:
       - source_labels: [__meta_kubernetes_pod_name]
         action: replace
         target_label: instance
+      # Sourcegraph specific customization. We want to add a label to every 
+      # metric that indicates the node it came from.
+      - source_labels: [__meta_kubernetes_endpoint_node_name]
+        action: replace
+        target_label: nodename
 
     # Example scrape config for probing services via the Blackbox Exporter.
     #
@@ -242,6 +247,11 @@ data:
       - source_labels: [__meta_kubernetes_namespace]
         action: replace
         target_label: ns
+      # Sourcegraph specific customization. We want to add a label to every 
+      # metric that indicates the node it came from.
+      - source_labels: [__meta_kubernetes_endpoint_node_name]
+        action: replace
+        target_label: nodename
 
       metric_relabel_configs:
       # cAdvisor-specific customization. Drop container metrics exported by cAdvisor
--- a/charts/sourcegraph/values.yaml
+++ b/charts/sourcegraph/values.yaml
@@ -704,8 +704,8 @@ nodeExporter:
     runAsUser: 65534
     runAsGroup: 65534
     readOnlyRootFilesystem: true
-  # -- Security context for the `node-exporter` container,
-  # learn more from the [Kubernetes documentation](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-container)
+  # -- Security context for the `node-exporter` pod,
+  # learn more from the [Kubernetes documentation](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-pod)
   podSecurityContext:
     fsGroup: 65534
     runAsGroup: 65534
--- a/charts/sourcegraph/values.yaml
+++ b/charts/sourcegraph/values.yaml
@@ -697,7 +697,7 @@ nodeExporter:
   extraArgs: []
   #   - --collector.diskstats.ignored-devices=^(ram|loop|fd|(h|s|v)d[a-z]|nvme\\d+n\\d+p)\\d+$
   #   - --collector.textfile.directory=/run/prometheus
-  # -- Security context for the `pgsql` container,
+  # -- Security context for the `node-exporter` container,
   # learn more from the [Kubernetes documentation](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-container)
   containerSecurityContext:
     allowPrivilegeEscalation: false
--- a/charts/sourcegraph/values.yaml
+++ b/charts/sourcegraph/values.yaml
@@ -661,6 +661,44 @@ openTelemetry:
       requests:
         cpu: "100m"
         memory: 100Mi
+
+nodeExporter:
+  # -- Enable `node-exporter`
+  enabled: true
+  image:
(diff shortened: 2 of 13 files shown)


charts/sourcegraph/templates/prometheus/prometheus.ConfigMap.yaml
```diff
@@ -171,6 +171,11 @@ data:
       - source_labels: [__meta_kubernetes_pod_name]
         action: replace
         target_label: instance
+      # Sourcegraph specific customization. We want to add a label to every
```
ggilmore, line 174: Here is the additional Prometheus label that I mentioned in the PR description.

charts/sourcegraph/values.yaml
```diff
@@ -661,6 +661,56 @@ openTelemetry:
       requests:
         cpu: "100m"
         memory: 100Mi
+
+nodeExporter:
+  # -- Enable `node-exporter`
+  enabled: true
+  image:
+    # -- Docker image tag for the `node-exporter` image
+    defaultTag: 179720_2022-10-25_4d925e87cfb8@sha256:2d9dcdf0b2226f0c3d550a64d2667710265462350a3ba9ebe37d0302bc64af0f
```
ggilmore, line 670: I used a valid image tab, but this isn't one that comes from a release (since haven't cut a release since node-exporter was introduced in https://github.com/sourcegraph/sourcegraph/pull/43107. Will the release tooling handle this special tag?

ggilmore:
cc @sourcegraph/delivery since I believe that you own the helm deployment

cc @sourcegraph/dev-experience since I believe that you're the long-term owners of our monitoring stack 

cc @sourcegraph/security, this is an example of how we're using the node-exporter image that I mentioned in https://sourcegraph.slack.com/archives/C1JH2BEHZ/p1666731965726319 (in case you have any concerns)

ggilmore:
Do I need to do anything else to announce this new service besides updating the changelog in this repo (as I did in this PR)?

charts/sourcegraph/values.yaml
slimsag: Seems fine either way. We should be able to check post-release and update then if needed.

slimsag (approved):
This looks good to me.

Nice work on the tests - and I think this being so similar to cadvisor already + the fact that you can disable it easily makes this quite low-risk.

Great work!

Issue #189, opened 2022-10-14
The test connection pod cannot be disabled or changed


uhthomas:
This is a major problem for many reasons:
- The templates cannot be rendered and applied with kubectl.
- The container image is always `busybox` from Docker Hub, with no ability to pin or change it.
- The templates cannot be applied in clusters which have strict registry enforcement (i.e cannot pull from Dockerhub).

Can this be removed or optionally disabled?

marcleblanc2:
Sorry for the delayed response, the connection test can now be disabled by overriding these values in your override file. 
https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/94a42b04c5e61837017471eab180e4238bb1d6af/charts/sourcegraph/values.yaml#L1355-L1357

PR #208, opened 2022-11-04
test-connection: allow test-connection pod configuration

jdpleiness:
Add support for disabling or changing the `test-connections` pod. See issue #189.

This PR makes the following changes:
* `testConnection` is enabled by default and uses the `busybox:latest` image from DockerHub.
* `testConnection` can be disabled or set to use a different image using the `override.yaml` file.

### Checklist

### Test plan
This is a change to a test. The manual testing process was followed as described above.

### DIFF
--- a/charts/sourcegraph/templates/tests/test-connection.yaml
+++ b/charts/sourcegraph/templates/tests/test-connection.yaml
@@ -1,15 +1,20 @@
+{{- if .Values.testConnection.enabled -}}
 apiVersion: v1
 kind: Pod
 metadata:
-  name: "sg-test-connection"
+  name: {{ .Values.testConnection.name }}
   labels:
     {{- include "sourcegraph.labels" . | nindent 4 }}
+    {{- if .Values.testConnection.labels }}
+      {{- toYaml .Values.testConnection.labels | nindent 4 }}
+    {{- end }}
   annotations:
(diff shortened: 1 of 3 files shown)


charts/sourcegraph/templates/tests/test-connection.yaml
```diff
@@ -1,15 +1,20 @@
...
+    {{- if .Values.testConnection.labels }}
+      {{- toYaml .Values.testConnection.labels | nindent 4 }}
+    {{- end }}
   annotations:
     "helm.sh/hook": test
 spec:
   containers:
     - name: wget
-      image: busybox
+      image: {{ .Values.testConnection.image.name }}:{{ .Values.testConnection.image.defaultTag }}
```
michaellzc, line 16: let's use the helper partial template like other deployment


```suggestion
      image: {{ include "sourcegraph.image" (list . "testConnection") }}
```

michaellzc (commented):
the CI is failling

would you run `./scripts/helm-docs.sh`?

let's also add some basic unit test https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/main/TEST.md

abeatrix (approved):
Other than the suggestion pointed out by Michael, the rest looks good to me! thanks jacob!

PR #209, opened 2022-11-08
prometheus: drop labels with duplicate nodename entries

ggilmore:
Follows https://github.com/sourcegraph/deploy-sourcegraph-dogfood-k8s/pull/5616, https://github.com/sourcegraph/deploy-sourcegraph-dogfood-k8s/pull/5613

This PR tweaks the Prometheus configuration to drop labels with an empty `nodename` parameter. 

Since https://github.com/sourcegraph/deploy-sourcegraph-helm/pull/194, there seems to be duplicate metrics (one with an empty `nodename`, one with a `populated` one). I'm unsure of the exact cause, but I can speculate that:
- perhaps the k8s api is returning empty responses in some cases
- perhaps the metric relabelling process ends up duplicating a time series 

This PR resolves this situtation by dropping the metrics with  empty `nodename` labels.

### Checklist

- [x] Update [changelog](https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/main/charts/sourcegraph/CHANGELOG.md) (not needed - change hasn't released yet)

### Test plan

Running on k8s.sgdev.org, see the following commits:

https://github.com/sourcegraph/deploy-sourcegraph-dogfood-k8s/pull/5613
https://github.com/sourcegraph/deploy-sourcegraph-dogfood-k8s/pull/5616

### DIFF
--- a/charts/sourcegraph/templates/prometheus/prometheus.ConfigMap.yaml
+++ b/charts/sourcegraph/templates/prometheus/prometheus.ConfigMap.yaml
@@ -254,7 +254,7 @@ data:
         target_label: ns
       # Sourcegraph specific customization. We want to add a label to every 
       # metric that indicates the node it came from.
-      - source_labels: [__meta_kubernetes_endpoint_node_name]
+      - source_labels: [__meta_kubernetes_pod_node_name]
         action: replace
         target_label: nodename
 
--- a/charts/sourcegraph/templates/prometheus/prometheus.ConfigMap.yaml
+++ b/charts/sourcegraph/templates/prometheus/prometheus.ConfigMap.yaml
@@ -176,6 +176,11 @@ data:
       - source_labels: [__meta_kubernetes_endpoint_node_name]
         action: replace
         target_label: nodename
+      metric_relabel_configs:
+      # Sourcegraph specific customization. Drop metrics with empty nodename responses from the k8s API
+      - source_labels: [nodename]
+        regex: ^$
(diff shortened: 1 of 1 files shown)


ggilmore:
cc @mucles When this gets merged, would you mind deploying this fix to the scaletesting cluster?

charts/sourcegraph/templates/prometheus/prometheus.ConfigMap.yaml
```diff
@@ -249,7 +254,7 @@ data:
         target_label: ns
       # Sourcegraph specific customization. We want to add a label to every 
       # metric that indicates the node it came from.
-      - source_labels: [__meta_kubernetes_endpoint_node_name]
+      - source_labels: [__meta_kubernetes_pod_node_name]
```
keegancsmith, line 257: this seems unrelated? what was the reason for this change?

keegancsmith (approved):

charts/sourcegraph/templates/prometheus/prometheus.ConfigMap.yaml
ggilmore: This was just a typo that existed in the original version of the prometheus node-exporter changes that I silently fixed here. `__meta_kubernetes_pod_node_name` is the correct name for this label (since we're scraping pods, not endpoints).  I figured this wasn't too bad since nobody has deployed this helm chart yet.

PR #245, opened 2023-02-20
otel: support custom `hostPort` values for agents

sanderginn:
Deploying the Helm chart multiple times on a single cluster will currently fail with OpenTelemetry enabled:
- The agent DaemonSet opens a `hostPort`, because services producing spans have to target the DaemonSet pod running on the same node as that service. This is achieved by setting `OTEL_EXPORTER_OTLP_ENDPOINT` to `hostIP` and then targeting the gRPC port (4317).
- A `hostPort` needs to be unique on each node.
- When deploying an additional instance of Sourcegraph with the Helm chart, it will try to deploy the DaemonSet, but the preexisting deployment already uses the `hostPort`, so the additional deployment fails.
- The existing implementation can only disable OpenTelemetry entirely, which drops tracing support.

This PR allows overriding the `hostPort`s of the agent DaemonSet, which will enable users to pick free ports on the node.

### Alternative approach
Theoretically, a single deployment of OpenTelemetry should suffice for each instance of Sourcegraph, but that would also mean a single configuration for both instances, which might not be desired (e.g. exporting to separate backends).  
This also requires a reasonable amount of refactoring, likely moving OpenTelemetry resources out of the `sourcegraph` Helm chart, as uninstalling one instance should not remove OpenTelemetry for any remaining instances.

### Checklist

### Test plan
- [x] unit tests pass
- [x] deploy on dogfood cluster with dogfood and scaletesting running in parallel

### DIFF
--- a/charts/sourcegraph/CHANGELOG.md
+++ b/charts/sourcegraph/CHANGELOG.md
@@ -7,7 +7,7 @@ Use `**BREAKING**:` to denote a breaking change
 <!-- START CHANGELOG -->
 
 ## Unreleased
-
+* The default `hostPort`s for OpenTelemetry agent pods deployed by the DaemonSet can now be overridden by setting `openTelemetry.agent.hostPorts` for one or more of `otlpGrpc`, `otlpHttp`, and `zpages`. This allows multiple instances of Sourcegraph to be deployed on a single cluster [#245](https://github.com/sourcegraph/deploy-sourcegraph-helm/pull/245).
 
 ## 4.5.0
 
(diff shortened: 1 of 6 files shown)


sanderginn:
* main
  * **PR #245** <a href="https://app.graphite.dev/github/pr/sourcegraph/deploy-sourcegraph-helm/245" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/sourcegraph/deploy-sourcegraph-helm/245?utm_source=stack-comment).

davejrt (approved):

PR #242, opened 2023-02-13
Update Searcher and Symbols to use StatefulSets

scjohns:
### Checklist

- [ ] Follow the [manual testing process](https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/main/TEST.md)
- [ ] Update [changelog](https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/main/charts/sourcegraph/CHANGELOG.md)

### Test plan

1. `helm lint charts/sourcegraph/.`
2. `helm template charts/sourcegraph`
3. Installed in minikube with no errors

PR #255, opened 2023-02-28
removed minReadySeconds settings from Searcher and Symbols

loujar:
This change is proposed because:
1. Symbols and Searcher were switched from Deployment Controllers to StatefulSets in [PR 242](https://github.com/sourcegraph/deploy-sourcegraph-helm/pull/242)
2. Kubernetes [introduced](https://kubernetes.io/blog/2021/08/27/minreadyseconds-statefulsets/) StatefulSet support for the `minReadySeconds` configuration in release v1.22
3. Sourcegraph docs advertise [support for Kubernetes 1.19+](https://docs.sourcegraph.com/admin/deploy/kubernetes#prerequisites), so we shouldn't ship default k8s configurations that are unsupported in 1.19

Alternative solutions could involve making `minReadySeconds` a configurable, optional helm value for these STSs, or possibly reverting these two services back to Deployments if we absolutely want to enforce this setting. The impact of removing this setting should be minimal, and we already ship other StatefulSets that do not define `minReadySeconds` setting.

### Checklist

### Test plan

1. `helm lint charts/sourcegraph/.`
2. `helm template charts/sourcegraph/ | grep -i "minReadySeconds" -B 2`
3. Deployed in GKE reference sandbox to confirm

### DIFF
--- a/charts/sourcegraph/templates/searcher/searcher.StatefulSet.yaml
+++ b/charts/sourcegraph/templates/searcher/searcher.StatefulSet.yaml
@@ -12,7 +12,6 @@ metadata:
     app.kubernetes.io/component: searcher
   name: {{ .Values.searcher.name }}
 spec:
-  minReadySeconds: 10
   replicas: {{ .Values.searcher.replicaCount }}
   revisionHistoryLimit: {{ .Values.sourcegraph.revisionHistoryLimit }}
   selector:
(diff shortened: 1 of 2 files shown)


scjohns (approved):
lgtm -- just add a test plan so all the checks pass

PR #276, opened 2023-04-28
Add optional PodDisruptionBudget for frontend

filiphaftek:
[issue](https://github.com/sourcegraph/customer/issues/2093)

Adds optional PodDisruptionBudget for frontend, gitserver and indexed-search.
For components with >1 replicas, it will increase SLA and stability, especially running on k8s as a service, where nodes are replaced by cloud operator.

Interesting reading:
[What if I don’t have pdb in my cluster?](https://innablr.com.au/blog/what-is-pod-disruption-budget-in-k8s-and-why-to-use-it/)

### Checklist

### Test plan

1. Unit tests
2. [e2e](https://github.com/sourcegraph/cloud/pull/1265) on internal instance


michaellzc:
nice!

my understanding of PDB is it’s only useful when your application is stateless or stateful with replication, and you’re trying to achieve HA.

yes, we should definitely add PDB for stateless services like frontend.

however, i don’t think it makes much sense to do so for any one of our statefulsets. they are all sharded not replicated, and we generally recommend running a single replica. For example, when running multiple gitserver replica, one of the shard goes down, we temporarily lose access to repos on it because they’re not replicated to else where at all.

Moreover, when running a single replica of statfulset and we have pdb configured, it might actually prevent node from being scaled down.

Therefore, I propose we remove PDB from all statefulesets. It’s just a temporary downtime we all have to live with.

filiphaftek:
> nice!
> 
> my understanding of PDB is it’s only useful when your application is stateless or stateful with replication, and you’re trying to achieve HA.
> 
> yes, we should definitely add PDB for stateless services like frontend.
> 
> however, i don’t think it makes much sense to do so for any one of our statefulsets. they are all sharded not replicated (unlikes etcd, consul), and we generally recommend running a single replica. For example, when running multiple gitserver replica, one of the shard goes down, we temporarily lose access to repos on it because they’re not replicated to else where at all.
> 
> Moreover, when running a single replica of statfulset and we have pdb configured, it might actually prevent node from being scaled down.
> 
> Therefore, I propose we remove PDB from all statefulesets. It’s just a temporary downtime we all have to live with.

Agree that for single replica this feature is useless. I only wonder what about having multiple (i.e. 2) sharded zoekts - ensure not all go down at once if node is being removed - wouldn't this help to provide better customer experience? 
@sourcegraph/search-team can you please add your feedback whether this might help or not?

keegancsmith:
We have partial availability when a zoekt or gitserver is unavailable. In the case of zoekt, we report back that there may be results missing. In the case of gitserver, operations for that repo will just fail loudly.

Partial availability sucks. To be honest the user experience that is better is to limit the time we have partial availability. IE to me it feels  better for 2 zoekts to be gone for 5min vs 2 zoekts taking turns being down over 10min.

Unless there are other implications of pod disruption budgets, I'm unsure how effective they will actually be at our scale.

PR #278, opened 2023-05-03
Jac/embeddings

jac:
Adds Embeddings service for Cody embeddings (disabled by default)
- SG blobstore + external storage supported by setting `.Values.embeddings.backend` to appropriate [values](https://docs.sourcegraph.com/cody/explanations/code_graph_context#storing-embedding-indexes)
  - Updates env vars on both the `embeddings` container and `worker` container.
  - Doesn't set on `frontend` despite being mentioned in docs as consensus is it's not actually needed on frontend (happy to be proven wrong)

### Checklist

### Test plan
Tested locally using blobstore + gcs (using `EMBEDDINGS_UPLOAD_GOOGLE_APPLICATION_CREDENTIALS_FILE_CONTENT`)
  - Successfully created embeddings for repositories

### DIFF
--- a/charts/sourcegraph/CHANGELOG.md
+++ b/charts/sourcegraph/CHANGELOG.md
@@ -8,6 +8,8 @@ Use `**BREAKING**:` to denote a breaking change
 
 ## Unreleased
 
+- Add Embeddings service (disabled by default) for Cody embeddings
+
 ## 5.0.1
 
 - Sourcegraph 5.0.1 is now available!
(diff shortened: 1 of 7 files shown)


eseliger (approved):

unknwon (approved):
Can't really review much but if it works then it's great, as long as it can be used to replaced the current Cloud-specific version (so we don't end up maintaining two charts of embeddings :D

PR #279, opened 2023-05-03
Executors: add native k8s executors

sanderginn:
## Description
This PR adds a chart for deploying native k8s executors. It also polishes the dind implementation.

### Checklist

### Test plan
- Ran the test steps as described in https://github.com/sourcegraph/sourcegraph/issues/50824
- Unit tests


sanderginn:
Here's an example `override.yaml` that I used to deploy executors with to the test cluster (in this case, a `batches` executor).

I created a dedicated node with the label `sourcegraph-executors=true` in order to target it with the node selector. This is required because the executor spawns a job that mounts the same PVC as the executor deployment, so they must reside on the same node. I will propose the addition of setting pod affinity on the jobs, because the current implementation doesn't support pain-free node autoscaling.

```yaml
executor:
  queueName: batches
  nodeSelector:
    sourcegraph-executors: "true"
  env:
    EXECUTOR_FRONTEND_URL:
      value: "http://sourcegraph-frontend:30080"
    # -- The shared secret configured in the Sourcegraph instance site config under executors.accessToken. Required.
    EXECUTOR_FRONTEND_PASSWORD:
      value: "hunter2hunter2hunter2hunter2"
    # -- The name of the queue to pull jobs from to. Possible values: batches and codeintel. Required.
    EXECUTOR_QUEUE_NAME:
      value: "batches"
    EXECUTOR_MAXIMUM_NUM_JOBS:
      value: "10"
    SRC_LOG_LEVEL:
      value: info
    SRC_LOG_FORMAT:
      value: condensed
    SRC_TRACE_LOG:
      value: "false"
    EXECUTOR_KUBERNETES_RESOURCE_REQUEST_MEMORY:
      value: 1Gi
    # For local deployments the host is 'host.docker.internal' and this needs to be true
    EXECUTOR_DOCKER_ADD_HOST_GATEWAY:
      value: "false"
    # Useful for debugging.
    EXECUTOR_KUBERNETES_NAMESPACE:
      value: sourcegraph
    KUBERNETES_KEEP_JOBS:
      value: "true"
    EXECUTOR_KEEP_WORKSPACES:
      value: "true"
```

charts/sourcegraph-executor/dind/values.yaml
```diff
@@ -52,68 +52,25 @@ storageClass:
...
-#   volumeName: ""
-#   # Use an existing configmap (specified by name) instead of creating one
-#   existingConfig: ""
-#   # Toggle deployment of applications on/off. Applies to databases and third-party applications
-#   enabled: true
-
 executor:
   enabled: true
   image:
     defaultTag: 5.0.6@sha256:f78a0ba3d84941ec3577f19075e935bec34a9bfaa9dcb85908142fff017e9d6d
```
Piszmog, line 58: Is this version controlled for the whole repo? Wondering since `EXECUTOR_QUEUE_NAMES` is introduced in 5.1.x

charts/sourcegraph-executor/k8s/templates/docker-daemon.ConfigMap.yaml
Piszmog, line 1: We don't need this for native k8s. Since native k8s interacts with the k8s api, executors do not do anything directly with docker registry

charts/sourcegraph-executor/k8s/templates/executor.Deployment.yaml
```diff
@@ -0,0 +1,116 @@
...
+      tolerations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.sourcegraph.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      volumes:
+        {{- if .Values.privateDockerRegistry.enabled }}
+        - name: docker-config
```
Piszmog, line 108: This can be removed

charts/sourcegraph-executor/k8s/templates/executor.Deployment.yaml
```diff
@@ -0,0 +1,116 @@
...
+          readinessProbe:
+            exec:
+              command:
+                - /usr/bin/pgrep
+                - -f
+                - /usr/local/bin/executor
+            periodSeconds: 5
+          volumeMounts:
+            {{- if .Values.privateDockerRegistry.enabled }}
+            - mountPath: /etc/docker/daemon.json
```
Piszmog, line 83: Remove

charts/sourcegraph-executor/k8s/templates/executor.Rolebinding.yaml
Piszmog, line 1: Should this be renamed `executor.RoleBinding.yaml` (big `B`)?

charts/sourcegraph-executor/k8s/templates/private-docker-registry.Deployment.yaml
Piszmog, line 1: don't need this for k8s

charts/sourcegraph-executor/k8s/templates/private-docker-registry.PersistentVolumeClaim.yaml
Piszmog, line 1: same

charts/sourcegraph-executor/k8s/templates/private-docker-registry.Service.yaml
Piszmog, line 1: same

Piszmog (commented):
Overall makes sense (first time I am seeing this syntax). The only change is that we do not need the private docker registry for the native k8s options. Since that doesn't do anything with docker directly but relies on kubernetes to do the creating of pods/containers

charts/sourcegraph-executor/README.md
```diff
@@ -1,102 +1,16 @@
...
-Check logs of the executor container
-
-```sh
-kubectl logs -l app=executor -c executor
-```
-
-[sourcegraph/sourcegraph]: ../sourcegraph/
-[sourcegraph/sourcegraph-executor]: ./
-[executor]: https://docs.sourcegraph.com/admin/executors
+## dind
```
danieldides, line 14: I think we should prefer "Docker in Docker (dind)" as the title. I'm not sure how common `dind` is for most people.

charts/sourcegraph-executor/dind/values.yaml
sanderginn: good catch. Updated to the same 5.1 image as the k8s chart. I think the images get updated during the release of a new chart version

sanderginn:
@Piszmog resolved the issues, I think it's a-okay for merging now

charts/sourcegraph-executor/README.md
```diff
@@ -1,102 +1,16 @@
...
-  DO NOT EDIT README.md directly.
-  README.md is automatically generated from README.md.gotmpl
--->
+# Sourcegraph executor helm charts
 
-# Sourcegraph Exexutor Helm Chart
+This directory contains two Helm charts for deploying executors on Kubernetes. [Read the docs](doc/admin/executors/deploy_executors_kubernetes.md) for more information on executors on Kubernetes.
 
-This chart contains two deployments, Sourcegraph Executors and a private Docker Registry. It is a supplemental chart for the parent [sourcegraph/sourcegraph] Helm Chart if you wish to deploy executors
+## k8s
```
Piszmog, line 5: Similar to the `dind`, maybe make this "Native Kubernetes"

Piszmog (approved):
Just one more minor change

sanderginn:
@jac thanks for that catch 👍 
I've added a separate step for packaging the executor charts.  

I've also changed the names of the charts. I've suffixed them with the implementation. This means that the `dind` chart, that used to be published as `sourcegraph-executor`, is now published as `sourcegraph-executor-dind`. Do you think this is an issue, as customers with the installed chart will have to use the new chart name to upgrade? (I don't know if there are many customers with this deployed, if any, tbh). I can revert the dind one to `sourcegraph-executor` if you think that's better.

PR #282, opened 2023-05-16
Embeddings Service Config

jac:
Changes how embeddings configuration is handled. This aligns more closely to how external blob storage is configured for code intel indexes

### Test plan
Testing by deploying locally; configured to connect to a GCS bucket. Confirmed embeddings uploaded correctly.


jdpleiness (approved):
Looks good! 🚀

PR #333, opened 2023-08-11
Update executors charts.yaml versions

jdpleiness:
Update executors versions in `Chart.yaml` files. These values seem to be missed by the release tooling.


sanderginn (approved):
Good catch! I'll file an issue.

PR #334, opened 2023-08-18
Qdrant: add helm charts

camdencheek:
This implements helm charts to add a Qdrant deployment to our cloud deployments. The intent is to replace the `embeddings` service entirely and to stop depending on blobstore for embeddings storage. 

### Checklist

### Test plan

I ran the cluster locally, ensuring that I can generate and search embeddings with Qdrant. 

I tested with the following `override.yaml`:
```yaml
# Disable SC creation
storageClass:
  create: false
  name: standard

sourcegraph:
  localDevMode: true
  image:
    defaultTag: insiders
    useGlobalTagAsDefault: true
# More values to be added in order to test your change

qdrant:
  enabled: true

gitserver:
  env:
    SRC_REPOS_DESIRED_PERCENT_FREE:
      value: "0"

frontend:
  env:
    QDRANT_ENDPOINT:
      value: "qdrant:6334"

```

### DIFF
--- a/charts/sourcegraph/templates/qdrant/qdrant.ConfigMap.yaml
+++ b/charts/sourcegraph/templates/qdrant/qdrant.ConfigMap.yaml
@@ -8,18 +8,24 @@ metadata:
     app.kubernetes.io/component: qdrant
 data:
   config.yaml: |
-    debug: true
-    log_level: INFO
+    debug: {{ .Values.qdrant.config.debug }}
+    log_level: {{ .Values.qdrant.config.log_level }}
     storage:
       storage_path: /data
       snapshots_path: /data/storage
       on_disk_payload: true
+    service:
+      http_port: 6333
+      grpc_port: 6334
+    telemetry_disabled: true
+    # The following parameters can be configured
+    # on a per-collection basis, so these are just defaults.
     performance:
       max_optimization_threads: 4
     optimizers:
       max_optimization_threads: 4
       mmap_threshold_kb: 1
-      indexing_threshold_kb: 0
+      indexing_threshold_kb: 0 # disable indexing
     hnsw_index:
       m: 8
       ef_construct: 100
@@ -27,8 +33,4 @@ data:
       max_indexing_threads: 4
       on_disk: true
       payload_m: 8
-    service:
-      http_port: 6333
-      grpc_port: 6334
-    telemetry_disabled: true
 {{- end }}
--- a/charts/sourcegraph/templates/qdrant/qdrant.ConfigMap.yaml
+++ b/charts/sourcegraph/templates/qdrant/qdrant.ConfigMap.yaml
@@ -12,6 +12,7 @@ data:
     log_level: INFO
     storage:
       storage_path: /data
+      snapshots_path: /data/storage
       on_disk_payload: true
     performance:
       max_optimization_threads: 4
--- a/charts/sourcegraph/templates/qdrant/qdrant.ConfigMap.yaml
+++ b/charts/sourcegraph/templates/qdrant/qdrant.ConfigMap.yaml
@@ -1,0 +1,33 @@
+{{- if .Values.qdrant.enabled -}}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Values.qdrant.name }}
+  labels:
+    deploy: sourcegraph
+    app.kubernetes.io/component: qdrant
+data:
+  config.yaml: |
+    debug: true
+    log_level: INFO
+    storage:
+      storage_path: /data
+      on_disk_payload: true
+    performance:
+      max_optimization_threads: 4
+    optimizers:
+      max_optimization_threads: 4
+      mmap_threshold_kb: 1
+      indexing_threshold_kb: 0
+    hnsw_index:
+      m: 8
+      ef_construct: 100
+      full_scan_threshold: 10
+      max_indexing_threads: 4
+      on_disk: true
+      payload_m: 8
+    service:
+      http_port: 6333
+      grpc_port: 6334
+    telemetry_disabled: true
+{{- end }}
--- a/charts/sourcegraph/templates/qdrant/qdrant.Service.yaml
+++ b/charts/sourcegraph/templates/qdrant/qdrant.Service.yaml
@@ -1,0 +1,29 @@
+{{- if .Values.embeddings.enabled -}}
+apiVersion: v1
+kind: Service
+metadata:
+  annotations:
+    prometheus.io/port: "6060"
+    sourcegraph.prometheus/scrape: "true"
+    {{- if .Values.frontend.serviceAnnotations }}
+    {{- toYaml .Values.frontend.serviceAnnotations | nindent 4 }}
+    {{- end }}
+  labels:
+    app: {{ .Values.embeddings.name }}
+    deploy: sourcegraph
+    app.kubernetes.io/component: embeddings
+    {{- if .Values.sourcegraph.serviceLabels }}
+      {{- toYaml .Values.sourcegraph.serviceLabels | nindent 4 }}
+    {{- end }}
+  name: {{ .Values.embeddings.name }}
+spec:
+  ports:
+  - name: http
+    port: 9991
+    protocol: TCP
+    targetPort: http
+  selector:
+    {{- include "sourcegraph.selectorLabels" . | nindent 4 }}
+    app: {{ .Values.embeddings.name }}
+  type: {{ .Values.embeddings.serviceType | default "ClusterIP" }}
+{{- end }}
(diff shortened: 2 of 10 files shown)


charts/sourcegraph/values.yaml
```diff
@@ -322,6 +322,52 @@ embeddings:
...
+  serviceAccount:
+    # -- Enable creation of ServiceAccount for `embeddings`
+    create: false
+    # -- Name of the ServiceAccount to be created or an existing ServiceAccount
+    name: ""
+    annotations: {}
+  extraVolumeMounts: {}
+  extraVolumes: {}
+  # -- PVC Storage Request for `qdrant` data volume
+  storageSize: 100Gi
```
camdencheek, line 369: How difficult is it to increase the size of a PVC?

charts/sourcegraph/values.yaml
```diff
@@ -322,6 +322,52 @@ embeddings:
...
+qdrant:
+  # -- Enable `qdrant`
+  enabled: false
+  # -- Name of the `qdrant` service
+  name: qdrant
+  image:
+    # -- Docker image name for the `embeddings` image
+    name: "qdrant"
+    # -- Docker image tag for the `embeddings` image
+    defaultTag: "239247_2023-08-18_5.1-433e1b1c997f@sha256:eafcd7af2aca699fa9c9ce8e6aa674cc0470441f794baf031296d5d1cdadd0bc"
```
camdencheek, line 334: Pointing to the latest build from main

charts/sourcegraph/values.yaml
```diff
@@ -322,6 +322,52 @@ embeddings:
   extraVolumeMounts: {}
   extraVolumes: {}
 
+qdrant:
+  # -- Enable `qdrant`
+  enabled: false
```
camdencheek, line 327: Disabled by default, just like embeddings

charts/sourcegraph/templates/qdrant/qdrant.ConfigMap.yaml
```diff
@@ -0,0 +1,34 @@
...
+      max_optimization_threads: 4
+      mmap_threshold_kb: 1
+      indexing_threshold_kb: 0
+    hnsw_index:
+      m: 8
+      ef_construct: 100
+      full_scan_threshold: 10
+      max_indexing_threads: 4
+      on_disk: true
+      payload_m: 8
```
camdencheek, line 29: Will probably pull these out to make them configurable, but they are just default values that can be overridden on collection creation which is handled in the application.

charts/sourcegraph/templates/qdrant/qdrant.Service.yaml
```diff
@@ -0,0 +1,31 @@
+{{- if .Values.qdrant.enabled -}}
+apiVersion: v1
+kind: Service
+metadata:
+  annotations:
+    sourcegraph.prometheus/scrape: "true"
+    prometheus.io/port: "6333"
```
camdencheek, line 7: I tested that these populate metrics in Grafana. I'll need to add a basic dashboard for qdrant

charts/sourcegraph/values.yaml
```diff
@@ -292,7 +292,7 @@ embeddings:
     name: "embeddings"
     # -- Docker image tag for the `embeddings` image
     defaultTag: "5.1.6@sha256:e849f52e38637882e5d2ba3d7d27a656d897c4b4e2905e1fdb843536d9c948ab"
-  # -- Resource requests & limits for the `worker` container,
+  # -- Resource requests & limits for the `embeddings` container,
```
camdencheek, line 295: unrelated, just fixing a couple of typos

charts/sourcegraph/values.yaml
```diff
@@ -322,6 +322,52 @@ embeddings:
...
+    defaultTag: "239247_2023-08-18_5.1-433e1b1c997f@sha256:eafcd7af2aca699fa9c9ce8e6aa674cc0470441f794baf031296d5d1cdadd0bc"
+  # -- Resource requests & limits for the `qdrant` container,
+  # learn more from the [Kubernetes documentation](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/)
+  resources:
+    limits:
+      cpu: "2"
+      memory: 8G
+    requests:
+      cpu: "500m"
+      memory: 2G
```
camdencheek, line 343: Are there any guidelines for sizing on cloud, other than just "whatever is needed for perf"?

charts/sourcegraph/values.yaml
camdencheek: It looks like we [allow volume expansion](https://sourcegraph.com/github.com/sourcegraph/deploy-sourcegraph-helm@7b2814d123c1739a29ebc0ede47097fe35608634/-/blob/charts/sourcegraph/templates/storageclass.yaml?L21-21) by default, but I'm honestly not sure what the ramificaitons are. 100Gi is large enough for a large chunk of customers, but certainly not all.

michaellzc: cloud prefers to start small and scale up as needed (scaling up is easy and we have monitoring for it)

charts/sourcegraph/templates/qdrant/qdrant.ConfigMap.yaml
michaellzc: let's add a note in the manifest and say they're basically unused

charts/sourcegraph/values.yaml
michaellzc: in cloud, we will provide our own override.

- https://github.com/sourcegraph/cloud/blob/main/overrides/global.override.yaml
- https://github.com/sourcegraph/cloud/blob/main/overrides/small-resources.override.yaml

generally start small and increase as needed

michaellzc (commented):
overall lgtm

question to @camdencheek 

- qdrant seems to support [cluster mode](https://github.com/qdrant/qdrant-helm/blob/ed31b29a39d6b2c841f9f14e439d5b53c5acc587/charts/qdrant/values.yaml#L104-L110). how does it work? do we need it to support higher traffic? it's okay to say this out of the scope for MVP.

question to @sourcegraph/team-release 

- how do we decide whether to use an upstream chart or roll our own? https://github.com/qdrant/qdrant-helm/

charts/sourcegraph/templates/qdrant/qdrant.ConfigMap.yaml
```diff
@@ -0,0 +1,34 @@
...
+    deploy: sourcegraph
+    app.kubernetes.io/component: qdrant
+data:
+  config.yaml: |
+    debug: true
+    log_level: INFO
+    storage:
+      storage_path: /data
+      snapshots_path: /data/storage
+      on_disk_payload: true
```
michaellzc, line 16: these seem to be global flags; we should parameterize them

camdencheek:
> do we need it to support higher traffic? 

I don't _think_ we'll need cluster mode. A single pod with 32GB memory and 4 cores was able to handle sourcegraph.com-level size and traffic in my tests. Most of the scaling benefit comes from being able to build an index rather than horizontal scaling.

It's possible that we'll want to support cluster mode in the future, but a single pod should be plenty for the MVP (and beyond).

charts/sourcegraph/values.yaml
camdencheek: got it, thanks!

camdencheek:
> how do we decide whether to use an upstream chart or roll our own? 

Oh, I kinda assumed that our chart shouldn't have dependencies and that we'd want to follow the same patterns across all our services. I guess that wasn't a clear assumption 😄

charts/sourcegraph/templates/qdrant/qdrant.ConfigMap.yaml
camdencheek: Do we need to parameterize them even if we don't want them to be configurable? We control the volume mount location as well, so I can't think of any reason to move the storage path elsewhere

michaellzc: at least `debug` should be configurable, no?

I'm not sure the implication of `debug: true` means for qdrant, but we can only expose some of the fields here.

camdencheek: Oooh, gotcha, yeah definitely. No idea why my eyes focused on just the storage path 😂

jdpleiness:
> > > how do we decide whether to use an upstream chart or roll our own?
> > 
> > Oh, I kinda assumed that our chart shouldn't have dependencies and that we'd want to follow the same patterns across all our services. I guess that wasn't a clear assumption 😄
> 
> 
> yeah, that's a question for https://github.com/orgs/sourcegraph/teams/team-release how they want to approach this in the future.
> 
> but it shoudln't be blocking us from merging this experimental thing.

Yeah, this is how we've been doing it, so it's fine for now 👍 

We're still working things like this over and deciding on the best way forward.

jdpleiness (approved):

michaellzc (approved):
👍

PR #336, opened 2023-08-22
Qdrant: use standard PersistentVolumeClaim

camdencheek:
As I was poking at this more with @ggilmore, we realized that the PVC was non standard in the helm charts, and instead we can use `volumeClaimTemplates` for qdrant like we do for the other services in the repository.

### Checklist

### Test plan

Ran through the same manual end-to-end tests as with the previous PR. Also restarted the qdrant service to ensure that the data was persisted across restarts.


michaellzc (approved):
👍

PR #367, opened 2023-10-19
revert #348 (Remove Github Proxy)

BolajiOlajide:
### Checklist

### Test plan

Reverting #348 as this PR shouldn't be merged until the next major / minor release.
We currently cherry-pick release artifacts from patch releases into the `main` branch and #348 isn't included in `5.2.x`

### DIFF
--- a/charts/sourcegraph/CHANGELOG.md
+++ b/charts/sourcegraph/CHANGELOG.md
@@ -8,11 +8,12 @@ Use `**BREAKING**:` to denote a breaking change
 
 ## Unreleased
 
+- The GitHub Proxy service has been removed and is no longer required. Remove all `githubProxy` fields from config, if set.
+
 
 ## 5.2.1
 
 - Sourcegraph 5.2.1 is now available
-- The GitHub Proxy service has been removed and is no longer required. Remove all `githubProxy` fields from config, if set.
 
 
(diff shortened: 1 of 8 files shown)


camdencheek (approved):

eseliger:
it would be okay to include the original PR in 5.2 if that makes it easier for you, the reason it wasn't removed it purely timing, not that it is still required in the 5.2 release

eseliger:
Also, can you explain this a bit more to me? (can do later if this is blocking rn)

We have a `release/5.2` branch, why don't we generate patch releases on top of that? Changing main without backporting anything breaking previous releases feels very odd to me.

camdencheek:
@eseliger that's good information, thanks! That actually makes this much easier. 

The thing we're running into is we usually backport the image hash updates from a patch release. I think the idea here is that `main` reflects the latest "recommended" set of images. However, now github proxy doesn't exist, so we're trying to figure out what to do about the fact that we have a github proxy release in 5.2.1, but not on main. To close this out, we were going to revert this PR so that we had an image to point to, but that's obviously a bit flawed.

Since the image is not actually needed now, I think it's fine to just merge the image updates back to main for all but github-proxy. Which means we can skip this revert.

PR #418, opened 2024-02-06
rfc-795 relase manifest: the skips are not needed anymore

burmudar:
The skips were added during testing early on

### Test plan
ran the release tooling locally without the skips and it succeeded


jhchabran (approved):

keegancsmith:
@burmudar This commit was added since the `release/5.3` branch was created. Does it need to be cherry-picked? Does it deserve a changelog entry?

burmudar:
@keegancsmith nope! This only affects the new release tooling and has no bearing on anything else.

PR #419, opened 2024-02-08
Update redis health checks based on dot-com

daxmc99:
From #inc-270

### Checklist

- [ ] Follow the [manual testing process](https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/main/TEST.md)
- [ ] Update [changelog](https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/main/charts/sourcegraph/CHANGELOG.md)
- [ ] Update [Kubernetes update doc](https://docs.sourcegraph.com/admin/updates/kubernetes)

### Test plan

We should try to match the config on dot-com and keep our learnings

PR #420, opened 2024-02-13
Leverage readiness script

daxmc99:
Use script from https://github.com/sourcegraph/sourcegraph/pull/60365
### Checklist

### Test plan

CI

### DIFF
--- a/charts/sourcegraph/templates/redis/redis-cache.Deployment.yaml
+++ b/charts/sourcegraph/templates/redis/redis-cache.Deployment.yaml
@@ -65,7 +65,7 @@ spec:
             command:
             - /bin/sh
             - -c
-            - readiness.sh
+            - /readiness.sh
         env:
         {{- range $name, $item := .Values.redisCache.env}}
         - name: {{ $name }}
--- a/charts/sourcegraph/templates/redis/redis-cache.Deployment.yaml
+++ b/charts/sourcegraph/templates/redis/redis-cache.Deployment.yaml
@@ -65,7 +65,7 @@ spec:
             command:
             - /bin/sh
             - -c
-            - /readiness.sh
+            - readiness.sh
         env:
         {{- range $name, $item := .Values.redisCache.env}}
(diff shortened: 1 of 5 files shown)


keegancsmith (approved):
if you plan on backporting this, can you move the changelog entry into the 5.3.0 section?

Edit: nevermind on the backport, see here https://github.com/sourcegraph/deploy-sourcegraph-helm/pull/419#issuecomment-1942062087

charts/sourcegraph/templates/redis/redis-cache.Deployment.yaml
```diff
@@ -65,15 +65,7 @@ spec:
...
-            - |
-              #!/bin/bash
-              response=$(
-                redis-cli ping
-              )
-              if [ "$response" != "PONG" ]; then
-                echo "$response"
-                exit 1
-              fi
+            - readiness.sh
```
eseliger, line 68: does this depend on the cwd? should we use `/readiness.sh` instead? Not a probe expert, so not sure how the WD is set

daxmc99: Updated! I think it uses the root but this is more explicit regardless

eseliger (approved):

daxmc99:
Hmm, I'm struggling to find this new script in our images
https://github.com/sourcegraph/sourcegraph/pull/60365#issuecomment-1942088509

daxmc99:
Works locally with this override.yaml file
```
# Disable SC creation
storageClass:
  create: false
  name: standard

sourcegraph:
  localDevMode: true
  image:
    defaultTag: "262142_2024-02-13_5.2-4db83b41b419"
    useGlobalTagAsDefault: true
# More values to be added in order to test your change

```

daxmc99:
@sourcegraph/release I updated the values.yaml to `262142_2024-02-13_5.2-4db83b41b419` 
Otherwise this change will prevent the current chart from starting

PR #434, opened 2024-03-21
Release/5.3

Chickensoupwithrice:
### Checklist

### Test plan
Release Image

### DIFF
--- a/charts/sourcegraph-executor/dind/Chart.yaml
+++ b/charts/sourcegraph-executor/dind/Chart.yaml
@@ -5,7 +5,7 @@ icon: https://sourcegraph.com/favicon.ico
 type: application
 
 # Chart version, separate from Sourcegraph
-version: "5.2.7"
+version: "5.3.0"
 
 # Version of Sourcegraph release
-appVersion: "5.2.7"
+appVersion: "5.3.0"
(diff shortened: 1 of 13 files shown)


Chickensoupwithrice:
Looks like releases haven't been merged into main since 5.3.0? Do we want to be doing this?

DaedalusG (approved):

BolajiOlajide (approved):
@jdpleiness Can you have a look at this? I know we normally don't cherry-pick the release branch changes to `main` here, but this is a better improvement for folks who are viewing our helm chart and don't know how the release branches work.

However, I wonder if this would break any changes to how the helm charts are uploaded. I know we once had an issue you helped fix but can't remember why that happened anymore, so just flagging in case there's potential for any breaking changes with this PR.

jdpleiness (approved):
@BolajiOlajide yeah this makes sense. The issue we face last time was that we had to patch a release version but the charts were already uploaded and our GH action doesn't overwrite an old chart.

In this case, the charts will get published to `https://helm.sourcegraph.com/insiders/5.3.x` rather than `https://helm.sourcegraph.com/release/x` which I don't see any issues with.

BolajiOlajide:
> @BolajiOlajide yeah this makes sense. The issue we face last time was that we had to patch a release version but the charts were already uploaded and our GH action doesn't overwrite an old chart.
> 
> In this case, the charts will get published to `https://helm.sourcegraph.com/insiders/5.3.x` rather than `https://helm.sourcegraph.com/release/x` which I don't see any issues with.

Thank you for double checking.

Chickensoupwithrice:
Since double checking that we don't need to merge this, I'm closing this one

PR #455, opened 2024-04-23
release: add check for gh auth status

BolajiOlajide:
### Checklist

### Test plan

Add step to check that release captain is authenticated via `gh` during the release process.

### DIFF
--- a/release.yaml
+++ b/release.yaml
@@ -13,6 +13,9 @@ requirements:
   - name: "GitHub CLI "
     cmd: "which gh"
     fixInstructions: "install GitHub cli"
+  - name: "GH auth status"
+    cmd: "gh auth status"
+    fixInstructions: "gh auth login"
   - name: "Docker username"
     env: "DOCKER_USERNAME"
   - name: "Docker password"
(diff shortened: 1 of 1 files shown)


jdpleiness (approved):
Nice!

DaedalusG (approved):
Nice I believe @Chickensoupwithrice ran into this

release.yaml
```diff
@@ -21,6 +21,9 @@ internal:
   create:
     steps:
       patch:
+          - name: "gh:auth"
+            cmd: |
+              gh auth status
```
jhchabran, line 26: I think we should instead move this to the `requirements` and instead call `checkDeps` in `(*releaseRunner) Test(context.Context) error`. The repetition here is very error prone.

jhchabran (changes requested):

release.yaml
jhchabran: https://github.com/sourcegraph/sourcegraph/pull/62143

jhchabran (approved):
LGTM, I'll fix the part where we need to run this in the tests, over https://github.com/sourcegraph/sourcegraph/pull/62143

PR #484, opened 2024-06-11
fix(ci): deploy-sourcegraph-helm doesn't have releases, just tags

Chickensoupwithrice:
Since we moved this action from the monorepo to here, we need to change it to match the semantics of this repo. 
The monorepo had "actual" github releases, but this repo only has tagged commits that contain the release artifact. 

This PR changes the action so that it runs on push of `v<VERSION_NUMBER>` instead of creation of a release. 

### Checklist

### Test plan

Run manually

### DIFF
--- a/.github/workflows/release-amis.yml
+++ b/.github/workflows/release-amis.yml
@@ -1,8 +1,9 @@
 name: build-ami
 
 on:
-  release:
-    types: [published]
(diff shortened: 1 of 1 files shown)


jdpleiness (approved):
Nice! Looks good to me 🚀

PR #494, opened 2024-07-03
al/appliance-add-namespace

Chickensoupwithrice:
### Checklist


charts/sourcegraph-appliance/templates/deployment.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: {{ include "sourcegraph-appliance.fullname" . }}
+  namespace: {{ .Values.namespace }}
```
craigfurman, line 5: I think it might be more idiomatic to use helm's own namespace features, rather than explicitly setting metadata.namespace. See https://github.com/sourcegraph/sourcegraph/pull/63669 - WDYT?

Chickensoupwithrice: Hmm, I think I'm more partial to the namespace living in the helm chart rather than being passed as a command line argument. I'm not sure how often someone would want to specify the namespace explicitly?

DaedalusG (approved):

charts/sourcegraph-appliance/templates/deployment.yaml
craigfurman: It's pretty common for chart authors to use helm's own namespace features AFAIK, including in our main helm chart. No metadata.namespace is set in https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/main/charts/sourcegraph/templates/gitserver/gitserver.StatefulSet.yaml and similar templates, for example. There is also no `namespace` value in that chart's values.yaml.

One issue with specifying metadata.namespace explicitly, is that helm has to decide which one wins if the admin passes `helm install --namespace`. It also requires chart authors to ensure that the namespace is configured correctly in every template, which is error-prone.

> I'm not sure how often someone would want to specify the namespace explicitly?

I do see what you mean, "sourcegraph" is a sensible default for sure!

Ultimately I'll leave it up to you though, we can make this approach work 😁

Chickensoupwithrice: You've convinced me that using helm's namespace is more idiomatic, and so I've decided to take on your feedback and implement this through that feature. :)

charts/sourcegraph-appliance/values.yaml
```diff
@@ -36,10 +35,12 @@ serviceAccount:
...
 podLabels: {}
 
-podSecurityContext: {}
+podSecurityContext:
+  {}
   # fsGroup: 2000
 
-securityContext: {}
+securityContext:
+  {}
```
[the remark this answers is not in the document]
Chickensoupwithrice: Yeah, `helm lint` iirc

charts/sourcegraph-appliance/templates/rolebinding.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
     app.kubernetes.io/component: sourcegraph-appliance
 subjects:
   - kind: ServiceAccount
-    name: {{ include "sourcegraph-appliance.serviceAccountName" . }}
+    name: {{ .Values.serviceAccount.name }}
```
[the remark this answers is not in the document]
Chickensoupwithrice: That sounds great! Thanks for the thorough review, much appreciated :)

charts/sourcegraph-appliance/values.yaml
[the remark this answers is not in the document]
craigfurman: Oh I forgot about that command! No worries

craigfurman (approved):
🚀

PR #495, opened 2024-07-03
feat(appliance): Add Maintaince UI

Chickensoupwithrice:
> [!NOTE]  
> This is part of a stack of PRs created using [git-machete](https://github.com/VirtusLab/git-machete)
> This is my first time using it
> The stack list includes:
> #494 - Add Namespace
> #495 - Maintaince UI  <- You are here
> #496 - Appliance Ingress
> #497 - Appliance Selfupdater

# Based on PR #494 

### Checklist

- [ ] Follow the [manual testing process](https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/main/TEST.md)
- [ ] Update [changelog](https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/main/charts/sourcegraph/CHANGELOG.md)
- [ ] Update [Kubernetes update doc](https://docs.sourcegraph.com/admin/updates/kubernetes)

### Test plan

Manual testing

PR #496, opened 2024-07-03
feat(appliance): Add Ingress and Service

Chickensoupwithrice:
> [!NOTE]  
> This is part of a stack of PRs created using [git-machete](https://github.com/VirtusLab/git-machete)
> This is my first time using it
> This stack includes:
> #494 - Add Namespace
> #495 - Maintaince UI 
> #496 - Appliance Ingress <- You are here
> #497 - Appliance Selfupdater

# Based on PR #495

This PR depends on https://github.com/sourcegraph/sourcegraph/pull/63627

### Checklist

### Test plan

Manual testing

### DIFF
--- a/charts/sourcegraph-appliance/templates/service.yaml
+++ b/charts/sourcegraph-appliance/templates/service.yaml
@@ -1,25 +1,26 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: {{ include "sourcegraph-appliance.fullname" . }}
-  {{- if .Values.frontend.serviceAnnotations }}
   annotations:
-    {{- toYaml .Values.frontend.serviceAnnotations | nindent 4 }}
-  {{- end }}
+    prometheus.io/port: "6060"
+    sourcegraph.prometheus/scrape: "true"
+    {{- if .Values.service.serviceAnnotations }}
+    {{- toYaml .Values.service.serviceAnnotations | nindent 4 }}
+    {{- end }}
   labels:
-    app: sourcegraph-frontend
+    app: sourcegraph-appliance
     deploy: sourcegraph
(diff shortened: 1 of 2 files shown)


charts/sourcegraph-appliance/templates/service.yaml
```diff
@@ -1,18 +1,26 @@
...
+    {{- end }}
   labels:
-    {{- include "sourcegraph-appliance.labels" . | nindent 4 }}
+    app: sourcegraph-appliance
+    deploy: sourcegraph
+    app.kubernetes.io/component: frontend
+    {{- if .Values.service.serviceLabels }}
+      {{- toYaml .Values.service.serviceLabels | nindent 4 }}
+    {{- end }}
+  name: sourcegraph-appliance
```
craigfurman, line 17: I'm a bit confused - shouldn't this be _frontend's_ service, but pointed at an appliance web app pod label query? I thought the plan was to do that, with the appliance reconfiguring this service to point at frontend once the backend is standing up, but maybe I'm out of date 🙏

Chickensoupwithrice: No, you're absolutely right, that is the desired outcome. But in order to get to Maintenance UI we still need a service and an ingress. I think you're right that the name should be `sourcegraph-frontend` instead of `sourcegraph-appliance` since it is frontend's service

craigfurman: Oh yeah for sure, sounds like we agree. To be clear I was suggesting that we rename these resources to be the same as those of frontend, then merge this.

craigfurman (approved):
🚀

DaedalusG (approved):

PR #498, opened 2024-07-05
fix/otel-jaeger: update jaeger deployment

DaedalusG:
### Checklist

### Test plan
Deploy locally and validate

### DIFF
--- a/charts/sourcegraph/templates/jaeger/jaeger.Deployment.yaml
+++ b/charts/sourcegraph/templates/jaeger/jaeger.Deployment.yaml
@@ -53,7 +53,7 @@ spec:
       - name: jaeger
         image: {{ include "sourcegraph.image" (list . "jaeger") }}
         imagePullPolicy: {{ .Values.sourcegraph.image.pullPolicy }}
-        args: {{- default (list "--memory.max-traces=20000", "--sampling.strategies-file=/etc/jaeger/sampling_strategies.json", "--collector.otlp.enabled") .Values.jaeger.args | toYaml | nindent 8 }}
+        args: {{- default (list "--memory.max-traces=20000" "--sampling.strategies-file=/etc/jaeger/sampling_strategies.json" "--collector.otlp.enabled") .Values.jaeger.args | toYaml | nindent 8 }}
         env:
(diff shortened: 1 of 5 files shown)


bobheadxi (approved):

charts/sourcegraph/templates/jaeger/jaeger.Deployment.yaml
```diff
@@ -53,7 +53,7 @@ spec:
       - name: jaeger
         image: {{ include "sourcegraph.image" (list . "jaeger") }}
         imagePullPolicy: {{ .Values.sourcegraph.image.pullPolicy }}
-        args: {{- default (list "--memory.max-traces=20000") .Values.jaeger.args | toYaml | nindent 8 }}
+        args: {{- default (list "--memory.max-traces=20000" "--sampling.strategies-file=/etc/jaeger/sampling_strategies.json" "--collector.otlp.enabled") .Values.jaeger.args | toYaml | nindent 8 }}
```
bobheadxi, line 56: Minor: does this mean if one provides their own args to the chart, they need to make sure to include the default args explicitly?

DaedalusG: Yeah it does, if they provide their own entry args these will be overwritten. Thanks for pointing it out, something for docs. Maybe worth it to add the args to the `values.yml` so they are recommended there as a default as well? cc: @michaellzc

michaellzc (approved):

Chickensoupwithrice (approved):

Chickensoupwithrice:
Tested deployment locally with kind, and was successful after a couple fixes :)

charts/sourcegraph/templates/otel-collector/otel-collector.Deployment.yaml
```diff
@@ -64,9 +64,9 @@ spec:
         - name: JAEGER_HOST
           value: jaeger-collector
         - name: JAEGER_OTLP_GRPC_PORT
-          value: 4320
+          value: "4320"
```
DaedalusG, line 67: dang, I wonder if I can get a linter that will detect this

PR #521, opened 2024-08-01
fix(appliance): chart version interpolation

craigfurman:
You can't interpolate in values.yaml, even with things set in Chart.yaml. The old code causes a literal `{{ .Chart.AppVersion }}` to appear in image tags, which is invalid.

### Checklist

### Test plan

E2E tested the appliance, including deploying SG, using the following command to deploy:

```
helm upgrade --install --namespace test \
  --set noResourceRestrictions=true \
  appliance ./charts/sourcegraph-appliance
```

### DIFF
--- a/charts/sourcegraph-appliance/values.yaml
+++ b/charts/sourcegraph-appliance/values.yaml
@@ -8,7 +8,7 @@ image:
   image: appliance
   pullPolicy: IfNotPresent
   # Overrides the image tag whose default is the chart appVersion.
-  tag: "{{ .Chart.AppVersion }}"
+  tag: &version "5.5.3738"
   # Version and Tag (above) are subtley different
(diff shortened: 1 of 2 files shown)


charts/sourcegraph-appliance/values.yaml
```diff
@@ -8,7 +8,7 @@ image:
   image: appliance
   pullPolicy: IfNotPresent
   # Overrides the image tag whose default is the chart appVersion.
-  tag: "{{ .Chart.AppVersion }}"
+  tag: &version "5.5.3738"
```
craigfurman, line 11: In the long run, https://linear.app/sourcegraph/issue/REL-322/appliance-helm-chart-should-be-versioned should replace this at release time.

nelsonjr (approved):

PR #526, opened 2024-08-02
Syntactic code intel worker deployment

antonsviridov-src:
Fixes GRAPH-786

This adds experimental syntactic worker to our helm charts.
The configuration is similar in structure to the precise code intel worker (and was copied from it) but has a few modifications, removing unused env variables.

Resource usage is so far copied from precise worker but it's not scientific - we'll adjust the ceiling and number of workers after testing on real instances.

### Checklist

### Test plan

- Tested manually on my local kind cluster

### DIFF
--- a/charts/sourcegraph/templates/syntactic-code-intel/worker.Deployment.yaml
+++ b/charts/sourcegraph/templates/syntactic-code-intel/worker.Deployment.yaml
@@ -62,8 +62,8 @@ spec:
           value: blobstore
         - name: SYNTACTIC_CODE_INTEL_UPLOAD_AWS_ENDPOINT
           value: http://blobstore:9000
-        - name: SYNTACTIC_CODE_INTEL_WORKER_PORT
-          value: "{{ .Values.syntacticCodeIntel.properties.workerPort }}"
+        - name: SYNTACTIC_CODE_INTEL_WORKER_ADDR
+          value: ":{{ .Values.syntacticCodeIntel.properties.workerPort }}"
         {{- end }}
         {{- include "sourcegraph.openTelemetryEnv" . | nindent 8 }}
         image: {{ include "sourcegraph.image" (list . "syntacticCodeIntel") }}
--- a/charts/sourcegraph/templates/syntactic-code-intel/worker.Deployment.yaml
+++ b/charts/sourcegraph/templates/syntactic-code-intel/worker.Deployment.yaml
@@ -1,0 +1,122 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  annotations:
+    description: Handles syntactic code intelligence indexing
+  labels:
+    {{- include "sourcegraph.labels" . | nindent 4 }}
+    {{- if .Values.syntacticCodeIntel.labels }}
+      {{- toYaml .Values.syntacticCodeIntel.labels | nindent 4 }}
+    {{- end }}
+    deploy: sourcegraph
+    app.kubernetes.io/component: syntactic-code-intel
+  name: {{ .Values.syntacticCodeIntel.name }}
+spec:
+  minReadySeconds: 10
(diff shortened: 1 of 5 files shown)


charts/sourcegraph/values.yaml
```diff
@@ -797,6 +797,46 @@ postgresExporter:
       cpu: 10m
       memory: 50Mi
 
+syntacticCodeIntel:
+  # -- Environment variables for the `syntactic-code-intel-worker` container
+  env:
+    NUM_WORKERS:
+      value: "1"
```
kritzcreek, line 804: What does this env variable control? We don't want to limit the concurrency of `scip-syntax` itself to 1, right?

antonsviridov-src: Frankly I'm unsure. I've only set this to 1 for the local testing with Kind (hence the draft PR), as I expect this will wreck my laptop if it controls number of container copies

antonsviridov-src: I've removed it, couldn't find any references of it anywhere in the codebase that weren't referring back to this usage..

charts/sourcegraph/templates/syntactic-code-intel/worker.Deployment.yaml
```diff
@@ -0,0 +1,126 @@
...
+    {{- include "sourcegraph.labels" . | nindent 4 }}
+    {{- if .Values.syntacticCodeIntel.labels }}
+      {{- toYaml .Values.syntacticCodeIntel.labels | nindent 4 }}
+    {{- end }}
+    deploy: sourcegraph
+    app.kubernetes.io/component: syntactic-code-intel
+  name: {{ .Values.syntacticCodeIntel.name }}
+spec:
+  minReadySeconds: 10
+  replicas: {{ .Values.syntacticCodeIntel.replicaCount }}
```
michaellzc, line 17: is the new code intel worker horizontally scalable? reason asking is `worker` appears to be but it's not.

will we see weird behavior if two instances are active at once? conflicted indexing results?

charts/sourcegraph/templates/syntactic-code-intel/worker.Service.yaml
```diff
@@ -0,0 +1,31 @@
...
+    app.kubernetes.io/component: syntactic-code-intel
+    {{- if .Values.syntacticCodeIntel.serviceLabels }}
+      {{- toYaml .Values.syntacticCodeIntel.serviceLabels | nindent 4 }}
+    {{- end }}
+  name: syntactic-code-intel-worker
+spec:
+  ports:
+  - name: http
+    port: {{ .Values.syntacticCodeIntel.properties.workerPort }}
+    targetPort: http
```
michaellzc, line 23: at a glance https://github.com/sourcegraph/sourcegraph/tree/main/cmd/syntactic-code-intel-worker/shared, it seems like it's pull-based, and doesn't expose any endpoint.

charts/sourcegraph/templates/syntactic-code-intel/worker.Deployment.yaml
antonsviridov-src: > is the new code intel worker horizontally scalable

It should be, yes. Similar to the precise codeintel worker, the identity of the worker instance is used to "lock" certain records during processing. As it's part of the design of dbworker (which we ultimately delegate to), I believe this shouldn't be a problem.

charts/sourcegraph/templates/syntactic-code-intel/worker.Service.yaml
antonsviridov-src: Well just like the precise codeintel worker, we expose a general health endpoint: https://github.com/sourcegraph/sourcegraph/blob/main/cmd/syntactic-code-intel-worker/shared/shared.go#L69-L76

antonsviridov-src:
@michaellzc do you have any more comments on this?

PR #559, opened 2024-10-07
sourcegraph: do not include frontend env var in migrator container

michaellzc:
this can result in duplicated env var and cause weird behaviour.

### Checklist

### Test plan

n/a

### DIFF
--- a/charts/sourcegraph/templates/frontend/sourcegraph-frontend.Deployment.yaml
+++ b/charts/sourcegraph/templates/frontend/sourcegraph-frontend.Deployment.yaml
@@ -59,10 +59,6 @@ spec:
         {{- include "sourcegraph.databaseAuth" (list . "codeIntelDB" "CODEINTEL_PG") | nindent 8 }}
         {{- include "sourcegraph.databaseAuth" (list . "codeInsightsDB" "CODEINSIGHTS_PG") | nindent 8 }}
         {{- end }}
-        {{- range $name, $item := .Values.frontend.env }}
-        - name: {{ $name }}
-          {{- $item | toYaml | nindent 10 }}
-        {{- end }}
         {{- range $name, $item := .Values.migrator.env }}
         - name: {{ $name }}
           {{- $item | toYaml | nindent 10 }}


michaellzc:
@eseliger this is what caused the broken `kustomize build | kubectl apply -f` when rolling out MT

charts/sourcegraph/templates/frontend/sourcegraph-frontend.Deployment.yaml
```diff
@@ -59,10 +59,6 @@ spec:
         {{- include "sourcegraph.databaseAuth" (list . "codeIntelDB" "CODEINTEL_PG") | nindent 8 }}
         {{- include "sourcegraph.databaseAuth" (list . "codeInsightsDB" "CODEINSIGHTS_PG") | nindent 8 }}
         {{- end }}
-        {{- range $name, $item := .Values.frontend.env }}
-        - name: {{ $name }}
-          {{- $item | toYaml | nindent 10 }}
-        {{- end }}
```
eseliger, line 65: I have no clue about this - are there any env vars in frontend that migrator would also need? I think in the MVU case it might need things like "where's the KMS secrets" etc 🤔 But not exactly sure there. I could see this being used at customers .. 🤔 

cc @sourcegraph/release

michaellzc: I think this is fine.

for db connection string, we don't read from `frontend.env`, it's always computed from the partial template

https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/fde5020096813f5da08b4226f82dd600b2731d03/charts/sourcegraph/templates/frontend/sourcegraph-frontend.Deployment.yaml#L52-L61

not sure about the KMS example?

DaedalusG: I'm concerned this might be used in autoupgrade -- reading 👀 _brain is cooked from the release stuff_

DaedalusG: Alright I think these is all we need:
```
      - PGHOST=pgsql
      - PGPORT=5432
      - PGUSER=sg
      - PGPASSWORD=sg
      - PGDATABASE=sg
      - PGSSLMODE=disable

      - CODEINTEL_PGHOST=codeintel-db
      - CODEINTEL_PGPORT=5432
      - CODEINTEL_PGUSER=sg
      - CODEINTEL_PGPASSWORD=sg
      - CODEINTEL_PGDATABASE=sg
      - CODEINTEL_PGSSLMODE=disable

      - CODEINSIGHTS_PGHOST=codeinsights-db
      - CODEINSIGHTS_PGPORT=5432
      - CODEINSIGHTS_PGUSER=postgres
      - CODEINSIGHTS_PGPASSWORD=password
      - CODEINSIGHTS_PGDATABASE=postgres
      - CODEINSIGHTS_PGSSLMODE=disable
```

And those should all be generated by this

DaedalusG (approved):

charts/sourcegraph/templates/frontend/sourcegraph-frontend.Deployment.yaml
michaellzc: yes, thank you

marcleblanc2:
This was a breaking change, which impacted self-hosted customers upgrading, which should have been communicated to them in advance of upgrading >= v5.8

PR #564, opened 2024-10-09
Adding pgsql datasource for Grafana

marcleblanc2:
For [PS-83: Create pgsql datasource in Grafana](https://linear.app/sourcegraph/issue/PS-83/create-pgsql-datasource-in-grafana)

### Checklist

### Test plan

Tested on test instance for upgrades and new installs, with and without the env vars configured

### Alternatives Considered
I tried to change the existing ConfigMap just mounted as a single file inside the directory, so then I could just add Postgres via a separate ConfigMap via subDir mount, but that threw Grafana for a loop because the datasources.yml file is not declarative (ex. Terraform), but just an input ingestion method to Grafana's config state (maintained in a MongoDB inside the Grafana container?), and when the datasources.yml file gets a new inode on an existing instance, Grafana fails to start up, because we have default:true on Prometheus, so Grafana seeing the same Prometheus data source in a new file at a new inode, so it errors out like we're trying to declare a second Prometheus datasource as also default, without first deleting the first one... we'd have to also migrate the Grafana config on existing instances by declaring a delete datasources block to delete the existing default Prometheus before creating a "new" one, and if we're going through all that, I'd say we should also set prune:true to cover us for the next time someone needs to change something, but that feature's not available in v7.5 :facepalm:

### DIFF
--- a/charts/sourcegraph/templates/grafana/grafana.ConfigMap.yaml
+++ b/charts/sourcegraph/templates/grafana/grafana.ConfigMap.yaml
@@ -15,6 +15,15 @@ data:
         type: Jaeger
         access: proxy
         url: http://{{ default "jaeger-query" .Values.jaeger.query.name }}:16686/-/debug/jaeger
+      - name: pgsql
+        type: postgres
+        url: $PGHOST:$PGPORT
+        user: $PGGRAFANAUSER
+        database: $PGDATABASE
+        secureJsonData:
+          password: $PGGRAFANAPASSWORD
+        jsonData:
+          sslmode: $PGSSLMODE
 kind: ConfigMap
 metadata:
   labels:


charts/sourcegraph/templates/grafana/grafana.ConfigMap.yaml
```diff
@@ -15,6 +15,15 @@ data:
...
         url: http://{{ default "jaeger-query" .Values.jaeger.query.name }}:16686/-/debug/jaeger
+      - name: pgsql
+        type: postgres
+        url: $PGHOST:$PGPORT
+        user: $PGGRAFANAUSER
+        database: $PGDATABASE
+        secureJsonData:
+          password: $PGGRAFANAPASSWORD
+        jsonData:
+          sslmode: $PGSSLMODE
```
jdpleiness, line 26: Can we make this optional with something like `{{- if .Values.grafana.postgres.enabled }}` wrapping it? 
I'd be worried about this being on by default.

marcleblanc2: Grafana doesn't create this datasource if the env vars are not defined, so the switch to enable it is defining all of these env vars on the Grafana container; it'd be redundant to add another config flag.

DaedalusG (approved):

charts/sourcegraph/templates/grafana/grafana.ConfigMap.yaml
DaedalusG: @jdpleiness had a call with @marcleblanc2 and discussed this, things look okay and yea grafana shouldn't deploy these unless the specific env vars are set. They're only set by default via our helpers in `frontend` and `migrator`

PR #580, opened 2024-11-18
Support migrator.databaseAuthOverrideEnvVars in sourcegraph-migrator chart job

loujar:
Allowing override of migrator auth ENVs (`.Values.migrator.databaseAuthOverrideEnvVars`) in the `sourcegraph-migrator` chart's `sourcegraph-migrator` job template, consistent with the `sourcegraph` chart's `sourcegraph-frontend` Deployment `migrator` initContainer
https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/v5.8.1579/charts/sourcegraph/templates/frontend/sourcegraph-frontend.Deployment.yaml#L57

### Checklist

### Test plan

CI


marcleblanc2 (approved):

DaedalusG:
I see why you're torn though, discussed with team a little and now I'm sympathetic to 1 since the frontend is already using the config defined in migrator, also yea, we should never really run migrator differently in different services (or at least I can't think of a case where we'd want that)

PR #632, opened 2025-02-03
feat: add runAsUser and runAsGroup to executor

Chickensoupwithrice:
The start of a resolution for [REL-690](https://linear.app/sourcegraph/issue/REL-690/add-runasuser-runasgroup-fsgroup-to-executor-helm-chart). 
Need to test this out to see how it works. There are sample overrides in [this thread](https://sourcegraph.slack.com/archives/C0418GDBT7S/p1738581012896529?thread_ts=1738580801.517539&cid=C0418GDBT7S)

### Checklist

### Test plan
Run the example overrides from the thread and see if the executor runs and doesn't fail. In particular we want a _named_ user that is not privileged.

### DIFF
--- a/charts/sourcegraph-executor/k8s/values.yaml
+++ b/charts/sourcegraph-executor/k8s/values.yaml
@@ -90,6 +90,12 @@ executor:
   namespace: "default"
   # -- The path to the kubeconfig file. If not specified, the in-cluster config is used.
   kubeconfigPath: ""
+  # -- The containerSecurityContext for the executor image
+  containerSecurityContext:
+    # @default -- `100`; accepts [0, 2147483647]
+    runAsUser: 100
+    # @default -- `101`; accepts [0, 2147483647]
+    runAsGroup: 101
   
   kubernetesJob:
     # -- The number of seconds after which a Kubernetes job will be terminated.
--- a/charts/sourcegraph-executor/k8s/values.yaml
+++ b/charts/sourcegraph-executor/k8s/values.yaml
@@ -91,11 +91,15 @@ executor:
   # -- The path to the kubeconfig file. If not specified, the in-cluster config is used.
   kubeconfigPath: ""
   # -- The containerSecurityContext for the executor image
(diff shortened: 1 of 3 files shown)


charts/sourcegraph-executor/k8s/values.yaml
```diff
@@ -90,6 +90,12 @@ executor:
   namespace: "default"
   # -- The path to the kubeconfig file. If not specified, the in-cluster config is used.
   kubeconfigPath: ""
+  # -- The containerSecurityContext for the executor image
+  containerSecurityContext:
+    # @default -- `100`; accepts [0, 2147483647]
+    runAsUser: 100
+    # @default -- `101`; accepts [0, 2147483647]
+    runAsGroup: 101
```
jdpleiness, line 98: Will adding this affect existing users?

marcleblanc2: Yeah, I'd suggest the default should not provide a value, just use the default already baked into the image, but provide the option for this to be configured.

marcleblanc2: While we're at it, should we also configure the `podSecurityContext` block as well? Quite similar, but also includes the `fsGroup` which is also be needed.

Would a `privileged:` config be relevant here?

Chickensoupwithrice: i'll add `fsGroup`, i think it is needed too

podSecurityContext takes precedence over containerSecurityContext if set, and `privileged` is only available for containerSecurityContext. This values section is a little misleading, it's actually setting the podSecurityContext, i'll fix it up

Chickensoupwithrice: we could set both, but seems unnecessary?

jdpleiness (approved):
@marcleblanc2 please run this through in your test cluster to make sure it meets the requirements.

Approving for now though as it looks good from here.

PR #666, opened 2025-04-10
 Fix Grafana's data source config for pgsql

marcleblanc2:
For Self-hosted Analytics project

### Checklist

### Test plan

Tested on self-hosted instance

### DIFF
--- a/charts/sourcegraph/templates/grafana/grafana.ConfigMap.yaml
+++ b/charts/sourcegraph/templates/grafana/grafana.ConfigMap.yaml
@@ -16,14 +16,15 @@ data:
         access: proxy
         url: http://{{ default "jaeger-query" .Values.jaeger.query.name }}:16686/-/debug/jaeger
       - name: pgsql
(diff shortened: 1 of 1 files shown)


jdpleiness (commented):
@marcleblanc2 can you  add a little more context here? I'm not sure I see the difference aside from the removal of `editable`

marcleblanc2:
> @marcleblanc2 can you add a little more context here? I'm not sure I see the difference aside from the removal of `editable`

Yes, adding `editable: true` to implement a Grafana-ism in Grafana's data source config file, as part of building Sourcegraph Analytics for our self-hosted / no-telemetry customers, so they can manage the database connection information, without requiring a bunch of manual maintenance on the Grafana dashboard JSON blob. This came up in my testing this week.

jdpleiness (approved):

jdpleiness:
I see. I know it's tempting to just "sort", but more often than not it's not really needed and makes reviewing what should be a simple change harder.

PR #683, opened 2025-05-22
Use Mise

burmudar:
Use mise in CI

### Test plan
CI

[_Created by Sourcegraph batch change `burmudar/use-mise`._](https://sourcegraph.sourcegraph.com/users/burmudar/batch-changes/use-mise)

### DIFF
--- a/.buildkite/hooks/asdf-pre-command
+++ b/.buildkite/hooks/asdf-pre-command
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -eu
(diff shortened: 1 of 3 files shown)


keegancsmith (approved):
just checking, so using mise is built into the stateless agent? How come falling back to asdf isn't?

burmudar:
> just checking, so using mise is built into the stateless agent? How come falling back to asdf isn't?

We do _kinda_. We fallback to installing asdf plugins - the old way.

The asdf install had a bit of a splitbrain install. The agent would do plugin installs, then the repository would do the tool install.

PR #752, opened 2025-10-02
Add support for deploying without using Kubernetes secrets

marcleblanc2:
...and to disable RBAC resources.

Customer's Kubernetes security policies block the creation of secrets and RBAC resources

### Checklist

### Test plan

Tested with customer

### DIFF
--- a/charts/sourcegraph/templates/_helpers.tpl
+++ b/charts/sourcegraph/templates/_helpers.tpl
@@ -45,7 +45,7 @@ For top-level services, pass in the top-level values:
 {{ include "sourcegraph.serviceAccountName" (list . "frontend") }}
 
 frontend:
-  serivceAccount:
+  serviceAccount:
     create: false
 
 For nested services, pass in the nested values:
@@ -249,6 +249,12 @@ app.kubernetes.io/name: jaeger
 {{- end }}
 
 {{- define "sourcegraph.redisConnection" -}}
+{{- if .Values.sourcegraph.disableKubernetesSecrets -}}
+- name: REDIS_CACHE_ENDPOINT
+  value: {{ .Values.sourcegraph.redisCacheEndpoint }}
+- name: REDIS_STORE_ENDPOINT
+  value: {{ .Values.sourcegraph.redisStoreEndpoint }}
+{{- else -}}
 - name: REDIS_CACHE_ENDPOINT
   valueFrom:
     secretKeyRef:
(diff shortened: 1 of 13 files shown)


charts/sourcegraph/templates/_helpers.tpl
```diff
@@ -249,6 +249,12 @@ app.kubernetes.io/name: jaeger
 {{- end }}
 
 {{- define "sourcegraph.redisConnection" -}}
+{{- if .Values.sourcegraph.disableKubernetesSecrets -}}
+- name: REDIS_CACHE_ENDPOINT
+  value: {{ .Values.sourcegraph.redisCacheEndpoint }}
+- name: REDIS_STORE_ENDPOINT
+  value: {{ .Values.sourcegraph.redisStoreEndpoint }}
+{{- else -}}
```
jdpleiness, line 257: This looks fine, but there's a case where someone disables secrets and forgets to set this, it's just going to return empty vars and fail at runtime.
We might want to add some validation here, possible something like:
```h
{{- define "sourcegraph.redisConnection" -}}
{{- if .Values.sourcegraph.disableKubernetesSecrets -}}
- name: REDIS_CACHE_ENDPOINT
  value: {{ required "sourcegraph.redisCacheEndpoint is required when sourcegraph.disableKubernetesSecrets is true" .Values.sourcegraph.redisCacheEndpoint }}
- name: REDIS_STORE_ENDPOINT
  value: {{ required "sourcegraph.redisStoreEndpoint is required when sourcegraph.disableKubernetesSecrets is true" .Values.sourcegraph.redisStoreEndpoint }}
{{- else -}}
- name: REDIS_CACHE_ENDPOINT
  valueFrom:
    secretKeyRef:
      key: endpoint
      name: {{ default .Values.redisCache.name .Values.redisCache.connection.existingSecret }}
- name: REDIS_STORE_ENDPOINT
  valueFrom:
    secretKeyRef:
      key: endpoint
      name: {{ default .Values.redisStore.name .Values.redisStore.connection.existingSecret }}
{{- end -}}
{{- end -}}
```

jdpleiness (commented):
Looks good overall. I think there is one minor issue around validation though that I commented on 🚀

charts/sourcegraph/templates/_helpers.tpl
marcleblanc2: Found a cleaner way to do this, re-using existing configs in the Helm chart:
```
.Values.redisCache.connection.endpoint
.Values.redisStore.connection.endpoint
```

PR #854, opened 2026-04-22
Add example override file for OpenShift

marcleblanc2:
### Checklist

Doc change.

### Test plan

Verify with customers.

### DIFF
--- a/charts/sourcegraph/examples/openshift/override.yaml
+++ b/charts/sourcegraph/examples/openshift/override.yaml
@@ -34,6 +34,9 @@
 #     OpenShift has a limit of 1 million files per volume
 #     so you need to scale gitserver and indexedSearch horizontally by incrementing replicaCount
 #     when they grow large enough to fail mounting the PV to a new host
+#     Create a free OpenShift developer account to see their docs:
+#     - https://access.redhat.com/solutions/7079407
+#     - https://access.redhat.com/solutions/6221251
 #
 #   Ingress / Router
 #
--- a/charts/sourcegraph/examples/openshift/override.yaml
+++ b/charts/sourcegraph/examples/openshift/override.yaml
@@ -11,8 +11,10 @@
 #   RBAC
 #     Some cluster security policies may block the Helm chart from creating RBAC resources,
 #     such as service accounts, roles, and role bindings.
-#     Sourcegraph only uses RBAC resources for Kubernetes service discovery, and
-#     workarounds are available if needed
+#     Sourcegraph only uses RBAC resources for Kubernetes service discovery, so that
+#     frontend and Prometheus can find how many replicas of gitserver, indexServer, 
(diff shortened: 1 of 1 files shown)


charts/sourcegraph/examples/openshift/override.yaml
```diff
@@ -0,0 +1,418 @@
...
+#     instead of letting this Helm chart create one
+#
+#   File count limits per PV
+#     OpenShift has a limit of 1 million files per volume
+#     so you need to scale gitserver and indexedSearch horizontally by incrementing replicaCount
+#     when they grow large enough to fail mounting the PV to a new host
+#
+#   Ingress / Router
+#
+#     HAProxy router timeout
```
trly, line 45: I'd also note to remind customers to check for other upstream proxies that may require overrides if just changing the HAProxy timeout doesn't resolve SSE issues.

charts/sourcegraph/examples/openshift/override.yaml
```diff
@@ -0,0 +1,418 @@
...
+#     Requires many configs to be specified explicitly
+#
+#   Image registry
+#     Most of our OpenShift customers use an internal image registry
+#
+#   StorageClass
+#     OpenShift instances often have their own StorageClasses already created,
+#     instead of letting this Helm chart create one
+#
+#   File count limits per PV
```
trly, line 35: https://access.redhat.com/solutions/6221251

There are a few workarounds, here. It seems like this is inherited from upstream k8s. Curious that we haven't seen it on other non-openshift deploys

trly (approved):
Looks good. My comments can be addressed if desired in the future.

charts/sourcegraph/examples/openshift/override.yaml
marcleblanc2: I remember this one from The Bank, before the docs got paywalled, it was due to RHOSCP's thing with UID/GID, every time it attaches a volume to a node, it runs chown UID:GID on all files in the PV, which timed out > 1 million files.

trly: There's a couple workarounds in that docs to somewhat mitigate it, as well as an option added in 4.16

if the volumes affected by the long SeLinux relabeling are accessed via the ReadWriteOnce (RWO) access mode, and the cluster is running OpenShift 4.16 or newer, consider first using [ReadWriteOncePod (RWOP) access mode in combination with the seLinuxMount option](https://access.redhat.com/solutions/7079407) instead of this solution if the CSI driver used supports it.

marcleblanc2: I've added the links in the notes. Thank you <3

charts/sourcegraph/examples/openshift/override.yaml
marcleblanc2: I've added the mention in the notes. Thank you <3

PR #865, opened 2026-05-11
Extend persistent volume configuration options

loujar:
Adding two additional PVC configuration options for the `sourcegraph/sourcegraph` helm chart for all stateful services:
- `storageSubPath`: allows mounting a subdirectory of a volume by setting the `subPath` on the primary data volume mount for each stateful service
- `storageAnnotations`: allows setting arbitrary annotations on PVC resources, for both standalone PersistentVolumeClaim resources and StatefulSet `volumeClaimTemplates` sections
### Checklist

### Test plan

CI

### DIFF
--- a/charts/sourcegraph/README.md
+++ b/charts/sourcegraph/README.md
@@ -41,6 +41,7 @@ In addition to the documented values, all services also support the following va
 | blobstore.serviceAccount.create | bool | `false` | Enable creation of ServiceAccount for `blobstore` |
 | blobstore.serviceAccount.name | string | `""` | Name of the ServiceAccount to be created or an existing ServiceAccount |
 | blobstore.storageSize | string | `"100Gi"` | PVC Storage Request for `blobstore` data volume |
+| blobstore.storageSubPath | string | `""` | Optional subPath for the `blobstore` primary data volume mount |
 | cadvisor.containerSecurityContext | object | `{"privileged":true}` | Security context for the `cadvisor` container, learn more from the [Kubernetes documentation](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-container) |
 | cadvisor.enabled | bool | `true` | Enable `cadvisor` |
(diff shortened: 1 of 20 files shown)


michaellzc (approved):
sure, but what is the use case? are they looking to share the same PVC instance across multiple services?

marcleblanc2:
LGTM. 

> what is the use case? are they looking to share the same PVC instance across multiple services?

I'll let Louis speak to the specifics of this case, but I can imagine that adding annotations to PVCs would be helpful with things like selecting which PVCs to include in snapshot policies (ex. gitserver yes, indexServer no), and subPath would help with mounting additional files from a configMap / secret, into the same destination dir.

As an aside, the subPath for Prometheus may be the workaround we need to support read-only root volumes for Prometheus, something I've got half sorted out.

marcleblanc2 (approved):

loujar:
> sure, but what is the use case? are they looking to share the same PVC instance across multiple services?

that's correct. This was a specific requirement that came up during a trial implementation. We worked around the constraints by manually templating & editing the helm chart resources to validate that we could support these requirements, but we want to manage this in standard helm chart configuration long term.

michaellzc:
> > sure, but what is the use case? are they looking to share the same PVC instance across multiple services?
> 
> that's correct. This was a specific requirement that came up during a trial implementation. We worked around the constraints by manually templating & editing the helm chart resources to validate that we could support these requirements, but we want to manage this in standard helm chart configuration long term.

not blocking. just curious on why? are they having constraint around number of PVCs they're allowed to provision? simplify backup of PVC (backing-up a single disk vs. many)?

DaedalusG (approved):

loujar:
> > > sure, but what is the use case? are they looking to share the same PVC instance across multiple services?
> > 
> > 
> > that's correct. This was a specific requirement that came up during a trial implementation. We worked around the constraints by manually templating & editing the helm chart resources to validate that we could support these requirements, but we want to manage this in standard helm chart configuration long term.
> 
> not blocking. just curious on why? are they having constraint around number of PVCs they're allowed to provision? simplify backup of PVC (backing-up a single disk vs. many)?

As I understand, this is just how this organization prefers to manage the volumes used by services in their shared cluster; a single volume shared across all associated PVCs used by the application. It's most likely related to their standardized DR procedures.

PR #885, opened 2026-06-09
feat: support per-service image repository override

michaellzc:
https://sourcegraph.slack.com/archives/C05MW2TMYAV/p1781020218336259

Adds a `<serviceName>.image.repository` value that overrides the global `sourcegraph.image.repository` for an individual service, resolved in the `sourcegraph.image` partial template and falling back to the global value when unset. The change is backward compatible — services without an override continue to use the global repository, and the override is scoped per image block (e.g. overriding `frontend.image.repository` does not move the frontend deployment's migrator init container). Includes documentation in `values.yaml`, the README, and the CHANGELOG, plus a 5-case helm-unittest suite covering default, per-service override, custom global repository, precedence, and isolation.

### Checklist

- [x] Update [Kubernetes update doc](https://docs.sourcegraph.com/admin/updates/kubernetes)

### Test plan

Added `charts/sourcegraph/tests/imageRepository_test.yaml`, a helm-unittest suite with 5 cases: (1) default uses the global repository, (2) a per-service `image.repository` redirects only that service, (3) a custom global repository is honored when no per-service value is set, (4) the per-service value wins over a custom global value, and (5) overriding one service leaves other containers (incl. the migrator init container) on the global repository. The full chart suite (97 tests) passes, confirming backward compatibility, and `./scripts/helm-docs.sh` produces no uncommitted changes.

### DIFF
--- a/charts/sourcegraph/CHANGELOG.md
+++ b/charts/sourcegraph/CHANGELOG.md
@@ -8,6 +8,7 @@ Use `**BREAKING**:` to denote a breaking change
 
 ## Unreleased
 
+- Added support for overriding the image repository on a per-service basis via `<serviceName>.image.repository`, falling back to the global `sourcegraph.image.repository` when unset
 - Added livenessProbe to zoekt-webserver in indexed-search to detect and restart hung pods
 - Fix Pod Disruption Budget for sourcegraph-frontend
 - Added a startup probe to the gitserver statefulset to give it time to run the on-disk migration from repo names to repo IDs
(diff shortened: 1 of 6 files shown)


michaellzc:
@DaedalusG release is not out yet, this should be included in 7.4 right?

DaedalusG:
@michaellzc the `7.4.x` branch is cut so it just needs the backport 👍🏼

DaedalusG (approved):

PR #582, opened 2024-11-20
feat(sourcegraph): add support to scale worker horizontally

michaellzc:
ref CLO-1023

https://sourcegraph.com/docs/admin/workers#3-split-jobs-and-scale-independently

`worker` is not really horizontally scalable, and it requires manually creating additionally `worker` deployments with the correct `WORKER_JOB_ALLOWLIST` and `WORKER_JOB_BLOCKLIST` environment variable.

This PR added native support for managing multiple worker deployment through helm template.

By default, it will work as is to day with a single `worker` deployment. To enable additional dedicated worker, user can specify the following overrides:

```yml
worker:
  replicas:
  - jobs:
    - workspaces-reconciler
  - jobs:
    - codeintel-upload-janitor
```

This will create 3 distinct worker deployment (`worker`, `worker-0`, and `worker-1`) with the correct `WORKER_JOB_*` env var. Additinoally, we also support overriding replica specific resources config. See unit test for more information. 

### Checklist

- [x] Follow the [manual testing process](https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/main/TEST.md)
- [ ] Update [changelog](https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/main/charts/sourcegraph/CHANGELOG.md)
- [x] Update [Kubernetes update doc](https://docs.sourcegraph.com/admin/updates/kubernetes)

### Test plan

CI, and manual testing with

```yml
worker:
  replicas:
  - jobs:
    - workspaces-reconciler
    resources:
      limits:
        cpu: 200m
        memory: 256Mi
      requests:
        cpu: 200m
        memory: 256Mi
```

```
 helm upgrade --install --create-namespace -n sourcegraph -f ./override.yaml sourcegraph charts/sourcegraph/.
```

all pods are healthy

CleanShot 2024-11-20 at 13 33 17

you can also play around with `helm template --debug -f ./override.yaml sourcegraph charts/sourcegraph/.` and inspect the rendered YAML

PR #887, opened 2026-06-10
feat(worker): support per-replica replicaCount override

michaellzc:
## Summary

Follow-up to #582, which added support for scaling `worker` horizontally by configuring dedicated replicas per set of jobs.

This PR adds a `replicaCount` override for each distinct worker replica. Previously every generated worker deployment (the primary `worker` and each dedicated `worker-N`) inherited the single top-level `worker.replicaCount`. Now each entry under `worker.replicas` can set its own `replicaCount`, falling back to `worker.replicaCount` when unset.

This lets operators scale individual dedicated workers horizontally only for jobs that are safe to scale that way.

```yaml
worker:
  replicas:
    - replicaCount: 3 # make sure jobs are safe to scale horizontally
      jobs:
        - job1
        - job2
    - jobs:
        - job3 # inherits worker.replicaCount
```

## Changes

- `templates/_worker.tpl`: added a `$replicaCount` parameter to the `sourcegraph.worker` helper and use it for `spec.replicas`.
- `templates/worker/worker.Deployment.yaml`: pass `worker.replicaCount` for the primary worker, and per-replica `replicaCount` (defaulting to `worker.replicaCount`) for each dedicated replica.
- `values.yaml`: documented the new per-replica `replicaCount` option.
- `tests/worker_test.yaml`: added tests covering the default-inheritance and per-replica override behaviour.
- `CHANGELOG.md`: added an entry.

## Test plan

- `helm unittest charts/sourcegraph` — all 99 tests pass (7 in the worker suite).
- `./scripts/helm-docs.sh` produces no uncommitted changes.

Slack thread: https://sourcegraph.slack.com/archives/C0ATB8N5P3Q/p1781106413632049

---
_Generated by [Claude Code](https://claude.ai/code/session_01L76QgBN8t9M3gbJo5NdyZo)_

### DIFF
--- a/charts/sourcegraph/CHANGELOG.md
+++ b/charts/sourcegraph/CHANGELOG.md
@@ -8,6 +8,7 @@ Use `**BREAKING**:` to denote a breaking change
 
 ## Unreleased
 
+- Added support for overriding `replicaCount` per dedicated `worker` replica via `worker.replicas[].replicaCount`, falling back to `worker.replicaCount` when unset
 - Added support for overriding the image repository on a per-service basis via `<serviceName>.image.repository`, falling back to the global `sourcegraph.image.repository` when unset
 - Added livenessProbe to zoekt-webserver in indexed-search to detect and restart hung pods
 - Fix Pod Disruption Budget for sourcegraph-frontend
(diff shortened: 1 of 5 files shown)


filiphaftek (approved):
Nice!

PR #921, opened 2026-08-07
gitserver: make PVC access modes configurable via storageAccessModes

marcleblanc2:
## Problem

The gitserver StatefulSet volumeClaimTemplate hardcodes `accessModes: [ReadWriteOnce]`.

On SELinux-enforcing nodes (Bottlerocket, EKS Auto Mode), every gitserver pod (re)start makes containerd recursively relabel every file on the repos volume with the new container's random MCS categories, because an RWO volume cannot use the `-o context` mount option (the `SELinuxMount` feature gate is off by default in Kubernetes <= 1.36, and managed offerings like EKS Auto Mode do not allow enabling it). On a repos volume with millions of git objects, this holds gitserver in `ContainerCreating` for 15-30+ minutes, with kubelet logging `failed to reserve container name ... another CreateContainer request is in progress` as its CRI calls time out and retry behind the still-running relabel walk.

## Fix

Add `gitserver.storageAccessModes`, defaulting to the previous hardcoded `["ReadWriteOnce"]` so existing deployments render identically.

Setting it to `["ReadWriteOncePod"]` enables the mount-option labeling path (`SELinuxMountReadWriteOncePod`, GA in Kubernetes 1.36): kubelet mounts the volume with `-o context=<label>` and the relabel walk is skipped entirely. Requires the pod to pin `gitserver.podSecurityContext.seLinuxOptions.level` and a CSI driver with `seLinuxMount: true` (the EBS CSI driver qualifies).

## Notes

- No behavior change for existing users: default renders the same manifest (verified with `helm template`; all 103 unit tests and 12 snapshots pass).
- Access modes are immutable on PVCs and `volumeClaimTemplates` are immutable on StatefulSets, so switching an existing deployment requires deleting the StatefulSet (`--cascade=orphan`) and recreating the PVC. Noted in the CHANGELOG.
- Diagnosed on an EKS Auto Mode cluster where a 3.7M-inode repos volume was fully relabeled (millions of xattr writes, ~9 GB of metadata IO) on every gitserver restart.

## Test plan

- `helm unittest charts/sourcegraph`: 24 suites, 103 tests, 12 snapshots pass
- `helm template` with defaults renders `ReadWriteOnce` (unchanged)
- `helm template --set 'gitserver.storageAccessModes={ReadWriteOncePod}'` renders `ReadWriteOncePod`
- README regenerated with helm-docs

### DIFF
--- a/charts/sourcegraph/CHANGELOG.md
+++ b/charts/sourcegraph/CHANGELOG.md
@@ -8,6 +8,7 @@ Use `**BREAKING**:` to denote a breaking change
 
 ## Unreleased
 
+- Added `gitserver.storageAccessModes` (default `["ReadWriteOnce"]`) to allow `["ReadWriteOncePod"]`, which lets Kubernetes mount the repos volume with `-o context` on SELinux-enforcing nodes (e.g. Bottlerocket / EKS Auto Mode) instead of recursively relabeling every file on each pod start. Changing this on an existing deployment requires recreating the StatefulSet and PVC, as both fields are immutable.
 - Added `searcher.autoCacheSize` (default `false`) to omit the `SEARCHER_CACHE_SIZE_MB` and `SYMBOLS_CACHE_SIZE_MB` env vars, letting `searcher` auto-size its cache to ~45% of the live cache volume so it tracks PVC expansion instead of staying frozen to the initial `storageSize`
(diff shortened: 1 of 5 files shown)


michaellzc (commented):
can you add a test to assert both default and override work?

marcleblanc2:
> can you add a test to assert both default and override work?

Done, thank you @michaellzc !

michaellzc (approved):

PR #940, opened 2026-09-04
gitserver: Expose HTTP clone listener

eseliger:
Exposes the Gitserver HTTP clone listener on port 3179 in the StatefulSet and headless Service that we're adding to have a dedicated endpoint for HTTP clone operations.

Also moves to use Gitserver's debug HTTP endpoints for startup and liveness probes which will cover both ports.
Also adds a readiness probe against `/ready`, which ideally prevents traffic being routed a bit too early.


filiphaftek (approved):
👍

